PR #37, opened 2022-02-26
Create & Delete Users

sean-smith:


app.py
```diff
@@ -109,6 +111,16 @@ def get_version_():
     def list_users_():
         return list_users()
 
+    @app.route("/manager/create_user",  methods=["POST"])
+    @authenticated("admin")
+    def create_user_():
+        return create_user()
+
+    @app.route("/manager/delete_user",  methods=["POST"])
```
charlesg3, line 119: This should technically be a `PUT` since it's not creating something new.

frontend/src/model.js
```diff
@@ -359,6 +359,42 @@ function ListUsers() {
...
+}
+
+function DeleteUser(user) {
+  var url = 'manager/delete_user';
+  request('post', url, user
+  ).then(response => {
+    if(response.status === 200) {
+      let returned_user = response.data;
+      console.log(`user ${returned_user.Username} deleted`);
+      clearState(['users', 'index', returned_user.Username], []);
```
charlesg3, line 387: extraneous `, []` at the end of this call -- you can just call `clearState(path)`

frontend/src/pages/Clusters/ClusterActions.js
```diff
@@ -64,8 +64,11 @@ export default function ClusterActions () {
       loadTemplate(jsyaml.load(configuration));
     });
   }
+
   const deleteCluster = () => {
-    setState(['app', 'clusters', 'clusterDelete', 'dialog'], true)
+    console.log(`Deleting: ${clusterName}`);
+    DeleteCluster(clusterName, (resp) => {DescribeCluster(clusterName); ListClusters()});
+    hideDialog('deleteCluster');
```
charlesg3, line 71: I think this call to `hideDialog` should probably go into the callback above (?)

frontend/src/pages/Clusters/ClusterActions.js
```diff
@@ -88,8 +91,10 @@ export default function ClusterActions () {
     GetDcvSession(instance.instanceId, user, callback);
   }
 
-  return <div style={{marginLeft: "20px"}}>
-    <ClusterDeleteDialog clusterName={clusterName} />
+  return (<><div style={{marginLeft: "20px"}}>
```
charlesg3, line 94: Because this component has a top-level `<div>` the `<>` and `</>` aren't needed here. If they're desired then the parenthesis can be removed and the div can be put on a newline (stylistically)

frontend/src/pages/Users/Users.js
```diff
@@ -11,16 +11,17 @@
 import React from 'react';
 import { useSelector } from 'react-redux'
 import { useCollection } from '@awsui/collection-hooks';
-import { setState, useState } from '../../store'
+import { clearState, setState, useState } from '../../store'
 
-import { ListUsers, SetUserRole } from '../../model'
+import { ListUsers, SetUserRole, CreateUser, DeleteUser } from '../../model'
```
charlesg3, line 16: alphabetize

frontend/src/pages/Users/Users.js
```diff
@@ -132,6 +145,11 @@ function UserList(props) {
           header: "Created",
           cell: item => <DateView date={item.UserCreateDate} /> || "-",
           sortingField: "UserCreateDate"
+        },
+        {
+          id: "delete",
```
charlesg3, line 150: I'd make this `id` be `actions` to be more extensible.

frontend/src/pages/Users/Users.js
```diff
@@ -132,6 +145,11 @@ function UserList(props) {
...
+          header: "Action",
+          cell: item => <Button className="action" onClick={() => {setState(['app', 'user', 'delete'], item); showDialog('deleteUser')}} >Delete</Button>|| "-",
```
charlesg3, line 152: Might factor this out into a component of `UserActions`

frontend/src/pages/Users/Users.js
```diff
@@ -145,17 +163,25 @@ function UserList(props) {
...
   const navigationOpen = useState(['app', 'sidebar', 'drawerOpen']);
+  const usernamePath = ['app', 'users', 'email'];
+  const username = useState(usernamePath);
   const refreshUsers = () => {
     ListUsers();
   }
 
+  const createUser = () => {
+    let user = {Username: username};
+    CreateUser(user);
```
charlesg3, line 181: Probably makes sense to put the `clearState` into a success callback for the `CreateUser` function. That way if there's an error (like the user already existing) then the input from the user doesn't get cleared out.

charlesg3 (changes requested):
A few very minor comments inlined. Otherwise, this look really nice. Good job on the extra effort to generalize the DeleteDialog into a component.

api/PclusterApiHandler.py
```diff
@@ -441,6 +441,23 @@ def list_users():
...
         return {"exception": str(e)}
 
+def delete_user():
+    try:
+        cognito = boto3.client("cognito-idp")
+        username = request.json.get("Username")
+        cognito.admin_delete_user(UserPoolId=USER_POOL_ID, Username=username)
+        return {'Username': username}
+    except Exception as e:
+        return {"exception": str(e)}
```
charlesg3, line 451: Most other errors are returning a `message` and this should return a tuple with a `500` as the second argument to indicate internal error.

sean-smith: Brilliant! Made that change and now you get a nice error message

image

frontend/src/model.js
```diff
@@ -359,6 +359,43 @@ function ListUsers() {
...
 
+function CreateUser(user) {
+  var url = 'manager/create_user';
+  request('put', url, user
+  ).then(response => {
+    if(response.status === 200) {
+      console.log("user added:", response.data);
+      let returned_user = response.data;
+      setState(['users', 'index', returned_user.Username], returned_user);
+      clearState(['app', 'users', 'email']);
```
charlesg3, line 370: Because this bit of state is related to UX (e.g. form input) it is generally kept out of this `model` namespace. The state manipulation in here largely deals with interactions with the back-end and putting (or clearing) the returned state / updates into the portion of the app-db that reflect what is provided by the underlying system.

For bits that refer to front-end-only logic, the management of that state is kept next to the components that use them. Typically the way this is done is by providing a `successCallback` and `errorCallback` as the last two arguments to this function, and then calling those callbacks (if they're non-nill ... e.g. `successCallback && successCallback(response.data)` and then in the component that makes the call here, it passes in an anonymous function (`() => {clearState(['app', 'users', 'email'])}`) to this function to perform the state update after the response is returned. That way the management of UX state doesn't leak into this model class which shouldn't have to care about those bits of data.

frontend/src/pages/Users/Users.js
charlesg3: The component will look like this:
```
function UserActions({ user }) {
  return <SpaceBetween direction="horizontal" size="s">
    <Button className="action" onClick={() => {setState(['app', 'user', 'delete'], user); showDialog('deleteUser')}} >Delete</Button>
  </SpaceBetween>
}
```

Using the component will change the line above to this:
```
cell: item => <UserActions user={item} />,
```

frontend/src/model.js
sean-smith: gotcha

frontend/src/pages/Users/Users.js
sean-smith: ok done

frontend/src/pages/Clusters/ClusterActions.js
charlesg3: Hmm... I think this should work as you provided. You can put a log in the anonymous function, but presumably the other functions are being called (e.g. `DescribeCluster` and `ListClusters`)  and `Delete success` is being printed?

frontend/src/pages/Users/Users.js
sean-smith: done

charlesg3 (approved):

PR #52, opened 2022-03-10
Cfnparams

bollig:
Added param groups and labels to clean up the CFN experience when deploying

### DIFF
--- a/infrastructure/pcluster-manager.yaml
+++ b/infrastructure/pcluster-manager.yaml
@@ -25,6 +25,31 @@ Parameters:
     Type: String
     Default: 3.1.2
 
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label: 
+          default: Admin User
+        Parameters: 
+          - EnableAuth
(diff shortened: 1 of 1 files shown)


infrastructure/pcluster-manager.yaml
```diff
@@ -25,6 +26,35 @@ Parameters:
...
+      - Label:
+          default: ParallelCluster API
+        Parameters:
+          - Version
+          - PublicEcrImageUri
+    ParameterLabels:
+      EnableMFA:
+        default: Require Multi-Factor Authentication for all Users
+      EnableAuth:
+        default: Enable Auth Roles (e.g., Admin, User, Guest)
```
sean-smith, line 51: This is actually:

```
When false this enables users to create an account & login. Only use this value for tutorials or short lived deployments.
```

bollig: Ok, I'll update. Is this long form the Description or you want it to be the Label?

sean-smith (approved):

PR #54, opened 2022-03-11
`[Docs]` Pcluster Manager Docs

sean-smith:


docs/content/01-getting-started/03-pcmanager-connect.md
```diff
@@ -0,0 +1,24 @@
++++
+title = "c. Connect to Pcluster Manager"
+weight = 13
++++
+
+1. Got to the AWS Console, in the search box search for [**AWS CloudFormation**](https://console.aws.amazon.com/cloudformation/home) and click on that service.
+
+2. You'll see a stack named **pcluster-manager**, click on that stack > **Outputs** Tab then click on the **PclusterManagerUrl** to connect.
```
charlesg3, line 8: The user should be able to use the url provided in the email here rather than going to outputs (which should simplify this step)...

Perhaps we should expand the email to make things totally obvious, and this step should just be "check your email and follow the instructions if you're logging in... and then keep these somewhere at the bottom for those doing workshops?

charlesg3 (approved):

docs/content/01-getting-started/03-pcmanager-connect.md
sean-smith: Agreed - I'll make the updates

PR #55, opened 2022-03-11
Support for environments without a default Vpc

maclema:
Allows passing a VPC and subnet ID to be used for the image build infrastructure to allow for launching in an environment where no default VPC exists.

Of note, the subnet passed in as a parameter should have internet access (i.e private subnet with NAT gateway routes, or a public subnet). When a public subnet is used, it must have MapPublicIpOnLaunch enabled as image builder won't automatically map a public IP address and the image build will timeout.

**In addition** - this also resolved an issue when deploying in accounts where SNS is still in sandbox. If MFA is false, SmsConfiguration can be ignored to prevent stack creation failure due to Cognito not being able to validate SMS ability of SNS.

Tests
Manually launched parallel cluster (via parallel cluster manager) stacks using a stack referencing stack set templates with these updates which exist in my personal AWS account S3 bucket.

### DIFF
--- a/infrastructure/pcluster-manager.yaml
+++ b/infrastructure/pcluster-manager.yaml
@@ -24,7 +24,19 @@ Parameters:
     Description: Version of AWS ParallelCluster to deploy
     Type: String
     Default: 3.1.2
+  ImageBuilderVpcId:
+    Description: Optional. Provide a specific vpc id to use for building the container images. Use this if you don't have a default vpc available.
+    Type: String
+  ImageBuilderSubnetId:
+    Description: Optional. Provide a specific subnet id to use for building the container images. Use this if you don't have a default vpc available.
+    Type: String
 
+Conditions:
+  NonDefaultVpc:
+    Fn::And:
+      - !Not [!Equals [!Ref ImageBuilderVpcId, ""]]
+      - !Not [!Equals [!Ref ImageBuilderSubnetId, ""]]
+    
 Mappings:
   PclusterManager:
     Constants:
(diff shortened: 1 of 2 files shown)


sean-smith:
Ok I've gone ahead and tested the default case i.e. no VPC/Subnet provided and can confirm that works.

A few thoughts:
1. Make the subnet / vpc the AWS-specific types `AWS::EC2::[Subnet/VPC]::Id`. This doesn't allow not specifying a subnet/vpc but I'm hoping the drop down would default to the default vpc (but would need to confirm this).

maclema:
Switched to use the AWS-specific parameter types. While they do not default to a vpc/subnet, I think this is a good option as if there isn't a VPC available it will prevent the stack from launching, and failing later.

infrastructure/pcluster-manager.yaml
```diff
@@ -25,6 +25,20 @@ Parameters:
...
+  ImageBuilderVpcId:
+    Description: Select the VPC to use for building the container images.
+    Type: AWS::EC2::VPC::Id
+    MinLength: 1 # Force this to be selected
+  ImageBuilderSubnetId:
+    Description: Select the subnet to use for building the container images.
+    Type: AWS::EC2::Subnet::Id
+    MinLength: 1 # Force this to be selected
+
+Conditions:
```
sean-smith, line 37: This will never be `""` now right?

sean-smith (commented):
Let's remove the condition and some of the conditional logic

infrastructure/pcluster-manager.yaml
maclema: Good call. Conditions removed. Running another launch test for sanity checks.

PR #56, opened 2022-03-12
Accounting documentation

sean-smith:


charlesg3:
Thanks for doing this! Some of the files that moved are not used anymore or were moved to be leveraged by the solution script.  Shouldn't be an issue though

charlesg3 (approved):

PR #62, opened 2022-03-14
`[Logs]` Use a table to allow searching.

charlesg3:
image


frontend/src/pages/Clusters/Logs.js
```diff
@@ -38,9 +49,101 @@ function LogEvents() {
...
+        empty: (
+          <EmptyState
+            title='No logs'
+            subtitle='No logs to display.'
+          />
+        ),
+        noMatch: (
+          <EmptyState
+            title='No matches'
+            subtitle='No logs match the filters.'
```
sean-smith, line 65: Add a note here that it doesn't search the entire log file.

sean-smith (approved):

PR #161, opened 2022-06-28
Set gp3 as default volume type

ermanno:
### Notes
This patch sets `gp3` as the default volume type.

### Tests
Tested locally by creating a cluster configuration with the Editor. Followed [these instructions](https://github.com/aws-samples/pcluster-manager#local-development) for local development.


frontend/src/pages/Configure/Components.js
```diff
@@ -525,7 +525,7 @@ function RootVolume({basePath, errorsPath}) {
       Volume Type:
       <Select
         disabled={editing}
-        placeholder="Default (gp2)"
+        placeholder="Default (gp3)"
```
charlesg3, line 528: I might be missing the context here -- but if the user doesn't actually select anything from this dropdown, then the value doesn't end up in the config -- is this correct?

If that's the case then we have a discrepancy -- without setting it explicitly then the true default depends on the version of PC (>3.2 it's gp3 otherwise gp2). So we can update this placeholder based on the PC version, or we can always be explicit and make gp3 be default regardless of the version of PC.

charlesg3: (It seems to be set explicitly in the EBS settings)

frontend/src/pages/Configure/Components.js
```diff
@@ -506,6 +506,11 @@ function RootVolume({basePath, errorsPath}) {
     clearEmptyNest(rootVolumeSizePath, 3);
   }
 
+  React.useEffect(() => {
+    if(rootVolumeType === null)
+      setState(rootVolumeTypePath, 'gp3');
```
charlesg3, line 511: nit: might bubble up `gp3` as a separate variable to avoid magic numbers

charlesg3 (approved):

PR #162, opened 2022-07-01
Cost Explorer

AndyHoNYC:
## Description

This adds a tab called "Cost" which summarizes Cluster cost by pulling in data from cost-explorer and graphing it by instance type. Also, this includes a button to activate cost allocation tags, which is a necessary pre-requisite for cost explorer.

## Changes

- Included new Cost.tsx file
- Added route `/manager/graph_data?cluster_name=Cluster&Start=Start&End=End/`
- Added route `/manager/activate_tags/`
- Added route `/manager/check_tag_status/`

## How Has This Been Tested?

- Tested locally using created clusters

https://user-images.githubusercontent.com/68314782/180088672-fad78396-854a-425f-92c8-7c8152289d64.mov

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -13,7 +13,6 @@ import json
 import os
 import re
 import time
-from types import TracebackType
 
 import boto3
 import botocore
@@ -43,11 +42,10 @@ AUTH_URL = os.getenv("AUTH_URL", f"{AUTH_PATH}/login")
 JWKS_URL = os.getenv("JWKS_URL")
 AUDIENCE = os.getenv("AUDIENCE")
 USER_ROLES_CLAIM = os.getenv("USER_ROLES_CLAIM", "cognito:groups")
-REDIRECT_URL = f"{SITE_URL}/login"
 PCLUSTER_COST_TAGS = ['parallelcluster:attributes', 
-        'parallelcluster:cluster-name', 'parallelcluster:compute-resource-name', 'parallelcluster:filesystem', 
-        'parallelcluster:networking', 'parallelcluster:node-type', 'parallelcluster:queue-name', 
-        'parallelcluster:resource', 'parallelcluster:version']
+    'parallelcluster:cluster-name', 'parallelcluster:compute-resource-name', 'parallelcluster:filesystem', 
+    'parallelcluster:networking', 'parallelcluster:node-type', 'parallelcluster:queue-name', 
+    'parallelcluster:resource', 'parallelcluster:version']
 
 try:
     if (not USER_POOL_ID or USER_POOL_ID == "") and SECRET_ID:
@@ -706,22 +704,22 @@ def activate_tags():
         cost_explorer = boto3.client('ce')
         newlist = [{'TagKey': tag, 'Status': 'Active'} for tag in PCLUSTER_COST_TAGS]
         cost_explorer.update_cost_allocation_tags_status(CostAllocationTagsStatus=newlist)
-        return "Activated Tags"
+        return {"message": "Activated Tags"}
     except Exception as e:
         return {"message": str(e)}, 500
-
+ 
 def get_graph_data():
     try:
         cluster_name = request.args.get("cluster_name")
-        Start = request.args.get("Start")
-        End = request.args.get("End")
-        if cluster_name is None or Start is None or End is None:
+        start = request.args.get("start")
+        end = request.args.get("end")
+        if cluster_name is None or start is None or end is None:
             return {"message": "cluster_name , start date, or end date are not specified"}, 400
         cost_explorer = boto3.client('ce')
         data = cost_explorer.get_cost_and_usage(
             TimePeriod={
-                'Start': Start,
-                'End': End
+                'Start': start,
+                'End': end
             },
             Filter={
                 'Tags': {
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -695,7 +695,7 @@ def _get_params(_request):
 
 def check_tags():
     try:
-        cost_explorer = boto3.client('ce',)
+        cost_explorer = boto3.client('ce')
         inactive_tags = cost_explorer.list_cost_allocation_tags(Status='Inactive',TagKeys=PCLUSTER_COST_TAGS)
         return {'TagStatus': not (len(inactive_tags['CostAllocationTags']) > 0)}
     except Exception as e:
@@ -715,7 +715,9 @@ def get_graph_data():
         cluster_name = request.args.get("cluster_name")
         Start = request.args.get("Start")
         End = request.args.get("End")
-        cost_explorer = boto3.client('ce',)
+        if cluster_name is None or Start is None or End is None:
+            return {"message": "cluster_name , start date, or end date are not specified"}, 400
+        cost_explorer = boto3.client('ce')
         data = cost_explorer.get_cost_and_usage(
             TimePeriod={
                 'Start': Start,
@@ -742,6 +744,18 @@ def get_graph_data():
                     'Key': 'INSTANCE_TYPE'
                 },
             ],
+        )
+        return data
+    except Exception as e:
+        return {"message": str(e)}, 500
+
+def get_budget_data():
+    try:
+        account_id = boto3.client('sts').get_caller_identity().get('Account')
+        cluster_name = request.args.get('cluster_name')
+        data = boto3.client('budgets').describe_budget(
+            AccountId=account_id,
+            BudgetName=cluster_name,
         )
         return data
     except Exception as e:
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -42,6 +42,16 @@ AUTH_URL = os.getenv("AUTH_URL", f"{AUTH_PATH}/login")
 JWKS_URL = os.getenv("JWKS_URL")
 AUDIENCE = os.getenv("AUDIENCE")
 USER_ROLES_CLAIM = os.getenv("USER_ROLES_CLAIM", "cognito:groups")
+REDIRECT_URL = f"{SITE_URL}/login"
+PCluster_tags = ['parallelcluster:attributes', 
+        'parallelcluster:cluster-name', 'parallelcluster:compute-resource-name', 'parallelcluster:filesystem', 
+        'parallelcluster:networking', 'parallelcluster:node-type', 'parallelcluster:queue-name', 
+        'parallelcluster:resource', 'parallelcluster:version']
+cost_explorer = boto3.client('ce',)
+
+
+
+
 
 try:
     if (not USER_POOL_ID or USER_POOL_ID == "") and SECRET_ID:
@@ -688,14 +698,9 @@ def _get_params(_request):
     return params
 
 # Check if parallelcluster tags are active
-def check_tags(self, inactive_tags):
+def check_tags():
     try:
-        cost_explorer = boto3.client('ce',)
-        tags = ['parallelcluster:attributes', 
-        'parallelcluster:cluster-name', 'parallelcluster:compute-resource-name', 'parallelcluster:filesystem', 
-        'parallelcluster:networking', 'parallelcluster:node-type', 'parallelcluster:queue-name', 
-        'parallelcluster:resource', 'parallelcluster:version']
-        inactive_tags = cost_explorer.list_cost_allocation_tags(Status='Inactive',TagKeys=tags)
+        inactive_tags = cost_explorer.list_cost_allocation_tags(Status='Inactive',TagKeys=PCluster_tags)
         if len(inactive_tags['CostAllocationtTags']) > 0:
             return True
         else:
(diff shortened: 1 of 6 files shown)


app.py
```diff
@@ -142,6 +142,19 @@ def delete_user_():
     @authenticated("admin")
     def set_user_role_():
         return set_user_role()
+    
+    # Andy
```
sean-smith, line 146: remove comment

app.py
```diff
@@ -142,6 +142,19 @@ def delete_user_():
...
+    @authenticated()
+    def check_tags():
+        return check_tags()
+
+    @app.route("/manager/activate_tags")
+    @authenticated()
+    def activate_tags():
+        return activate_tags()
+
+    # End
```
sean-smith, line 157: remove

api/PclusterApiHandler.py
```diff
@@ -663,9 +663,64 @@ def _get_params(_request):
     params.pop("path")
     return params
 
+# Check if parallelcluster tags are active
+def check_tags(self, inactive_tags):
+    try:
+        cost_explorer = boto3.client('ce',)
+        tags = ['parallelcluster:attributes',
```
sean-smith, line 670: Let's pull these tags out into a global variable defined at the top and re-use that in both functions.

api/PclusterApiHandler.py
```diff
@@ -663,9 +663,64 @@ def _get_params(_request):
...
+        return {"message": str(e)}, 500
 
-# Proxy
-
+# Activate parallelcluster tags in cost explorer
+def activate_tags(self, cost_explorer):
+    try:
+        updated_tags = cost_explorer.update_cost_allocation_tags_status(
+            CostAllocationTagsStatus=[
+                {
```
sean-smith, line 687: use global variable here

api/PclusterApiHandler.py
```diff
@@ -33,6 +33,15 @@
 SECRET_ID = os.getenv("SECRET_ID")
 ENABLE_MFA = os.getenv("ENABLE_MFA")
 SITE_URL = os.getenv("SITE_URL", API_BASE_URL)
+PCluster_tags = ['parallelcluster:attributes',
```
sean-smith, line 36: caps for global var `PCLUSTER_COST_TAGS`

api/PclusterApiHandler.py
```diff
@@ -33,6 +33,15 @@
 SECRET_ID = os.getenv("SECRET_ID")
 ENABLE_MFA = os.getenv("ENABLE_MFA")
 SITE_URL = os.getenv("SITE_URL", API_BASE_URL)
+PCluster_tags = ['parallelcluster:attributes', 
+        'parallelcluster:cluster-name', 'parallelcluster:compute-resource-name', 'parallelcluster:filesystem', 
+        'parallelcluster:networking', 'parallelcluster:node-type', 'parallelcluster:queue-name', 
+        'parallelcluster:resource', 'parallelcluster:version']
+cost_explorer = boto3.client('ce',)
```
sean-smith, line 40: Let's keep this in the function

api/PclusterApiHandler.py
```diff
@@ -33,6 +33,15 @@
...
+
```
sean-smith, line 41: too much whitespace. Remove

api/PclusterApiHandler.py
```diff
@@ -680,44 +684,10 @@ def check_tags(self, inactive_tags):
...
-                    'Status': 'Active'
-                },
-                {
-                    'TagKey': 'parallelcluster:version',
-                    'Status': 'Active'
-                }
-            ]
-        )
+        for tag in PCluster_tags:
+            updated_tags = cost_explorer.CostAllocationTagStatus.append({'Tagkey': tag, 'Value': 'Active'})
```
sean-smith, line 690: We still need to call the API. I was just referring to the way we pass in the list of tags. 

Let's use a python list comprehension here: https://www.w3schools.com/python/python_lists_comprehension.asp

api/PclusterApiHandler.py
```diff
@@ -663,9 +672,25 @@ def _get_params(_request):
     params.pop("path")
     return params
 
+# Check if parallelcluster tags are active
+def check_tags():
+    try:
+        inactive_tags = cost_explorer.list_cost_allocation_tags(Status='Inactive',TagKeys=PCluster_tags)
+        if len(inactive_tags['CostAllocationtTags']) > 0:
```
sean-smith, line 679: Single line return

app.py
```diff
@@ -142,6 +142,16 @@ def delete_user_():
     @authenticated("admin")
     def set_user_role_():
         return set_user_role()
+    
+    @app.route("/manager/check_tag_status", methods=["GET"])
```
sean-smith, line 146: remove `methods=["GET"]`

api/PclusterApiHandler.py
```diff
@@ -685,9 +690,59 @@ def _get_params(_request):
...
+        return "Activated Tags"
+    except Exception as e:
+        return {"message": str(e)}, 500
 
+def get_graph_data():
+    try:
+        cluster_name = request.args.get("cluster_name")
+        Start = request.args.get("Start")
+        End = request.args.get("End")
+        cost_explorer = boto3.client('ce',)
```
BarcoMasile, line 715: Nitpic(non-blocking):
you could remove the extra comma as it's not needed

api/PclusterApiHandler.py
```diff
@@ -685,9 +690,59 @@ def _get_params(_request):
...
+        cost_explorer.update_cost_allocation_tags_status(CostAllocationTagsStatus=newlist)
+        return "Activated Tags"
+    except Exception as e:
+        return {"message": str(e)}, 500
 
+def get_graph_data():
+    try:
+        cluster_name = request.args.get("cluster_name")
+        Start = request.args.get("Start")
+        End = request.args.get("End")
```
BarcoMasile, line 714: Suggestion:
the only thing I would add here is a check to make sure all these three values are actually provided
if any of these is `None`, than I would return an error 400 bad request since the request would be improper

api/PclusterApiHandler.py
```diff
@@ -685,9 +690,59 @@ def _get_params(_request):
...
+        newlist = [{'TagKey': tag, 'Status': 'Active'} for tag in PCLUSTER_COST_TAGS]
+        cost_explorer.update_cost_allocation_tags_status(CostAllocationTagsStatus=newlist)
+        return "Activated Tags"
+    except Exception as e:
+        return {"message": str(e)}, 500
 
+def get_graph_data():
+    try:
+        cluster_name = request.args.get("cluster_name")
+        Start = request.args.get("Start")
```
charlesg3, line 713: even though the parameters are upper-case in the call, variables are lower (snake) case in python unless you are defining a class by convention.

api/PclusterApiHandler.py
```diff
@@ -685,9 +690,59 @@ def _get_params(_request):
...
+    except Exception as e:
+        return {"message": str(e)}, 500
 
-# Proxy
+def activate_tags():
+    try:
+        cost_explorer = boto3.client('ce')
+        newlist = [{'TagKey': tag, 'Status': 'Active'} for tag in PCLUSTER_COST_TAGS]
+        cost_explorer.update_cost_allocation_tags_status(CostAllocationTagsStatus=newlist)
+        return "Activated Tags"
```
charlesg3, line 706: I'd suggest returning the same shape here as you do when there's an error, so that the parsing is consistent on the frontend.

frontend/src/old-pages/Clusters/Cost.tsx
```diff
@@ -0,0 +1,293 @@
...
+          variant="h2"
+          actions={
+            <SpaceBetween
+              direction="horizontal"
+              size="xs"
+            >
+              <DateRangePicker
+              onChange={({ detail }) => {fetchGraphData(detail.value); setDateValue(detail.value); }}
+                value={dateValue}
+                relativeOptions={[
```
charlesg3, line 245: I'd factor these options out as the actual view here is particularly long and this is just data

frontend/src/old-pages/Clusters/Cost.tsx
```diff
@@ -0,0 +1,293 @@
...
+                  startTimeLabel: "Start time",
+                  endTimeLabel: "End time",
+                  clearButtonLabel: "Clear and dismiss",
+                  cancelButtonLabel: "Cancel",
+                  applyButtonLabel: "Apply"
+                }}
+                placeholder="Filter by a date range" />
+        <Button disabled={tag_active} onClick={activate_tags} variant="primary">Activate Tags</Button> 
+        <HelpTooltip> 
+        The NoInstanceType category includes miscellaneous costs such as EBS, read more about <a href="https://aws.amazon.com/ebs/pricing/" target="-blank"> EBS Pricing</a>.
```
charlesg3: This should be internationalized -- here's an example of where this was done previously: https://github.com/aws-samples/pcluster-manager/pull/218

frontend/src/old-pages/Clusters/Cost.tsx
```diff
@@ -0,0 +1,293 @@
...
+        </HelpTooltip>
+            </SpaceBetween>
+          }
+        >
+          Cluster Costs
+        </Header>
+      }
+    >
+         <BarChart
+        series={stack_info}
```
charlesg3: Indentation looks off here -- perhaps run this through a prettier?

frontend/src/old-pages/Clusters/Cost.tsx
```diff
@@ -0,0 +1,293 @@
...
+                  startDateLabel: "Start date",
+                  endDateLabel: "End date",
+                  startTimeLabel: "Start time",
+                  endTimeLabel: "End time",
+                  clearButtonLabel: "Clear and dismiss",
+                  cancelButtonLabel: "Cancel",
+                  applyButtonLabel: "Apply"
+                }}
+                placeholder="Filter by a date range" />
+        <Button disabled={tag_active} onClick={activate_tags} variant="primary">Activate Tags</Button>
```
charlesg3: This tab is cluster specific, however the activation of tags seems to be account specific? If so then this probably belongs somewhere else (a new settings page?)

frontend/src/old-pages/Clusters/Cost.tsx
```diff
@@ -0,0 +1,293 @@
...
+    setState(['app', 'cost', cluster_name, 'max'], max);
+    }
+    GetGraphData(query_costs, cluster_name, Start, End)
+  }
+
+  const fetchGraphData = (val: any) => {
+    let start_value = ""
+    let end_value = ""
+    if (!val) {
+      let current_date = new Date();
```
charlesg3, line 187: javascript variables should be camelCase

frontend/src/old-pages/Clusters/Cost.tsx
```diff
@@ -0,0 +1,293 @@
...
+              const day_data = data_by_instance_type?.map((obj2: any) => {
+                if (obj2.Keys[0] == type) {
+                    total = Number(obj2.Metrics.NetUnblendedCost.Amount) + total
+                    day_info.push({
+                      x: new Date(obj.TimePeriod.Start),
+                      y: Number(parseFloat(obj2.Metrics.NetUnblendedCost.Amount).toFixed(2)),
+                    },)
+                }
+              })
+              if (total > max) {
```
charlesg3, line 138: Is it possible to make this a one-liner with: `max = Math.max(max, total);` ?

frontend/src/old-pages/Clusters/Cost.tsx
```diff
@@ -0,0 +1,293 @@
...
+      const usage_data = data['ResultsByTime'];
+      const dates: Date[] = [];
+      const instance_types = new Set();
+      const stack_info: any[] = [];
+      let day_info: any[] = [];
+      let total = 0;
+      let max = 0;
+  
+      // Find XDomain
+      const x_domain = usage_data?.map((obj: any) => {
```
charlesg3, line 110: obj is a bit generic, is it possible to come up with a more descriptive name here?

frontend/src/old-pages/Clusters/Cost.tsx
```diff
@@ -0,0 +1,293 @@
...
+          dates.push(new Date(obj.TimePeriod.Start));
+      })
+    
+      // Find All Instance Types
+      const data_instance_types = usage_data?.map((obj: any) => {
```
charlesg3, line 115: here you are assigning the result of a map to a variable but never using it and then performing operations within the map -- this is an odd mix of imperative and functional programming.

I would suggest either only using the result of the `map` function (or converting to a reduce if that makes more sense) or use an imperative structure like `for`

charlesg3 (changes requested):
comments inlined

frontend/src/old-pages/Clusters/Cost.tsx
AndyHoNYC: This is outside of the scope of the project. Since activating tags is a one-time thing and there's no cost associated with it, it's a safe operation for any user (not just cluster admins) to do. For organizational purposes makes more sense on a settings page but again a bit outside of scope.

api/PclusterApiHandler.py
```diff
@@ -685,12 +691,76 @@ def _get_params(_request):
...
+        cost_explorer = boto3.client('ce')
+        newlist = [{'TagKey': tag, 'Status': 'Active'} for tag in PCLUSTER_COST_TAGS]
+        cost_explorer.update_cost_allocation_tags_status(CostAllocationTagsStatus=newlist)
+        return {"message": "Activated Tags"}
+    except Exception as e:
+        return {"message": str(e)}, 500
+ 
+def get_graph_data():
+    try:
+        cluster_name = request.args.get("cluster_name")
```
charlesg3, line 713: nit: I'd move these 3 assignments and the check / validation above the try as they shouldn't throw and it decreases the amount of the function that is wrapped in the try (which makes it easier to read as the checks are at the top).

frontend/src/old-pages/Clusters/Cost.tsx
```diff
@@ -0,0 +1,418 @@
...
+                href="https://us-east-1.console.aws.amazon.com/billing/home?#/budgets/overview"
+                target="_blank"
+              >
+                {" "}
+                AWS Budgets
+              </a>
+              .
+            </HelpTooltip>
+            <Button
+              href={`https://${region}.console.aws.amazon.com/cost-management/home?region=${region}"#/custom?groupBy=InstanceType&hasBlended=false&hasAmortized=false&excludeDiscounts=true&excludeTaggedResources=false&excludeCategorizedResources=false&excludeForecast=false&timeRangeOption=Custom&granularity=Daily&reportName=&reportType=CostUsage&isTemplate=true&filter=${JSON.stringify(
```
charlesg3, line 307: I'd break this out into a separate variable as it's encoding a bit of logic that would make the view easier to read if factored out.

charlesg3 (changes requested):
Two minor comments inlined.

frontend/src/old-pages/Clusters/Cost.tsx
AndyHoNYC: resolved

api/PclusterApiHandler.py
AndyHoNYC: resolved

AndyHoNYC:
Rebased off origin/main.

charlesg3:
It looks like there's a few typing errors to be fixed (seen in the `frontend-tests`)

charlesg3 (dismissed):

PR #165, opened 2022-07-04
Add ONTAP/OpenZFS support

mtfranchetto:
## Notes
This PR adds support for [FSx ONTAP](https://aws.amazon.com/fsx/netapp-ontap/) and [FSx OpenZFS](https://aws.amazon.com/fsx/openzfs/).

### Changelist
- existing filesystem are filtered to apply only Lustre FS to the corresponding storage type, fixing a bug where Windows File Server FS were selectable
- return FSx volumes from the API
- add support for ONTAP/OpenZFS depending on the ParallelCluster version (3.2.0 and above)
- fixed a bug with EBS: linking an existing EBS volume wasn't working with the wizard

## Testing
- created a new cluster with OpenZFS/ONTAP volumes attached
- verified that the new ZFS/ONTAP filesystem do not appear under the Lustre section
- verified that the Lustre storage is still working
- linked an existing EBS volume and verified that the wizard is working

https://user-images.githubusercontent.com/3091286/177185857-bffb1f5f-7f6f-45d9-81e9-7940e88134fb.mov

### DIFF
--- a/frontend/src/pages/Configure/Storage.js
+++ b/frontend/src/pages/Configure/Storage.js
@@ -449,6 +449,11 @@ function EbsSettings({index}) {
   )
 }
 
+/*
+ * Used to configure the Storage part of the Cluster wizard:
+ *  - controls whether a storage type can be created or just linked
+ *  - specify if a storage type can be mounted as a file system or just one of its volumes
+ */
 const STORAGE_TYPE_PROPS = {
   "FsxLustre": {
     canCreate: true,
--- a/frontend/src/pages/Configure/Storage.js
+++ b/frontend/src/pages/Configure/Storage.js
@@ -460,18 +460,9 @@ function StorageInstance({index}) {
   const existingId = useState(existingPath) || "";
   const storages = useState(storagePath);
 
-  const fsxFilesystems = useState(['aws', 'fsx_filesystems']) || [];
+  const fsxFilesystems = useState(['aws', 'fsxFilesystems']);
   const efsFilesystems = useState(['aws', 'efs_filesystems']) || [];
   const editing = useState(['app', 'wizard', 'editing']);
-
-  const fsxName = (fsx) => {
-    var tags = fsx.Tags;
-    if(!tags) {
-      return null;
-    }
-    tags = fsx.Tags.filter((t) => {return t.Key === "Name"})
-    return (tags.length > 0) ? tags[0].Value : null
-  }
 
   const removeStorage = (type) => {
     if(index === 0 && storages.length === 1)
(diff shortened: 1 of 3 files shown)


frontend/src/pages/Configure/Storage.js
```diff
@@ -448,31 +449,51 @@ function EbsSettings({index}) {
   )
 }
 
+/*
+ * Used to configure the Storage part of the Cluster wizard:
+ *  - controls whether a storage type can be created or just linked
+ *  - specify if a storage type can be mounted as a file system or just one of its volumes
+ */
+const STORAGE_TYPE_PROPS = {
```
ermanno, line 457: Since we have introduced this, we should use `STORAGE_TYPE_PROPS[storageType].canCreate` in [Filesystems.js](https://github.com/aws-samples/pcluster-manager/blob/main/frontend/src/pages/Clusters/Filesystems.js#L31-L32) as well.

frontend/src/pages/Configure/Storage.js
```diff
@@ -542,11 +533,12 @@ function StorageInstance({index}) {
                         value={existingId}
                         onChange={(({detail}) => {setState(existingPath, detail.value)})} />
                     </div>,
-                  "FsxLustre": <FormField label="FSx Filesystem">
+                  "FsxLustre": <FormField label="FSx Lustre Filesystem">
                     <Select
                       placeholder="Please Select"
                       selectedOption={existingId && idToOption(existingId)} label="FSx Filesystem" onChange={({detail}) => {setState(existingPath, detail.selectedOption.value)}}
-                      options={fsxFilesystems.map((x, i) => {return {value: x.FileSystemId, label: (x.FileSystemId + (fsxName(x) ? ` (${fsxName(x)})` : ""))}})}
+                      options={fsxFilesystems.lustre.map((fs) => {return {value: fs.id, label: fs.name}})}
```
mendaomn, line 570: issue: As far as I can tell, we can have `fs.name === null` but we got rid of the `FileSystemID` as part of the label. Can't we end up in a situation where we have an empty label?

mtfranchetto: Good point, I'll restore the original behavior, and I'll keep the same also for OPENZFS and ONTAP

frontend/src/pages/Configure/Storage.js
mtfranchetto: We'll keep these kind of refactoring for the near future (when converting to Typescript) when we can then make storage types really "pluggable"

frontend/src/pages/Configure/Storage.js
mtfranchetto: Fixed!

mendaomn (approved):

PR #172, opened 2022-07-07
feat: make user roles claim in the jwt configurable, using 'cognito:groups' as default

tmscarla:
#### Context
In the context of implementing a configurable generic OIDC provider in current codebase, we want to get rid of the hardcoded dependencies of `cognito:groups` and introduce a configurable parameter for the name of the user groups claim in the JWT.

#### Solution
This PR aims to be a starting point in the direction of having a configurable OIDC provider in the ParallelCluster Manager. 

On the backend side, a new environment variable `USER_GROUPS_CLAIM` was introduced, with `cognito:groups` as default. On the frontend, the occurrences of `cognito:groups` were removed in favor of a generic `user_groups`, which is set in the `get_identity` function in the backend.

### Out of Scope
* Make the parameter effectively configurable through CloudFormation template. The overall UX setup will be handled in a dedicated story.

### Testing
* Tested frontend locally
* Tested backend deploying lambda through `update.sh` script:

```
{"attributes":{"email":"tsscarla@amazon.com","phone_number":"+10000000000","sub":"6343e9cb-32a3-42aa-92ae-400a1f5181f1"},"auth_time":1657208843,"client_id":"115p8h8vnrhp9sfrcieggbqdcq","event_id":"bee73da5-dfb4-4c31-baa6-1423829414eb","exp":1657295243,"iat":1657208843,"iss":"https://cognito-idp.eu-west-1.amazonaws.com/eu-west-1_nJFByWJpL","jti":"5e3925f6-bc4a-479e-9359-0df170e00ca2","origin_jti":"3db9f5a0-3b4a-4970-b128-019fd0bf18f7","scope":"openid email","sub":"6343e9cb-32a3-42aa-92ae-400a1f5181f1","token_use":"access","user_roles":["user","admin"],"username":"6343e9cb-32a3-42aa-92ae-400a1f5181f1","version":2}
```

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -33,6 +33,7 @@ CLIENT_SECRET = os.getenv("CLIENT_SECRET")
 SECRET_ID = os.getenv("SECRET_ID")
 ENABLE_MFA = os.getenv("ENABLE_MFA")
 SITE_URL = os.getenv("SITE_URL", API_BASE_URL)
+USER_ROLES_CLAIM = os.getenv("USER_ROLES_CLAIM")
 
 try:
     if (not USER_POOL_ID or USER_POOL_ID == "") and SECRET_ID:
@@ -137,7 +138,7 @@ def authenticate(group):
         return auth_redirect()
     except jose.exceptions.JWSSignatureError:
         return logout()
-    if not disable_auth() and (group != "guest") and (group not in set(decoded.get("cognito:groups", []))):
+    if not disable_auth() and (group != "guest") and (group not in set(decoded.get(USER_ROLES_CLAIM, []))):
         return auth_redirect()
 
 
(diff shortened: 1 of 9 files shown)


frontend/src/store.js
```diff
@@ -143,12 +143,12 @@ function useState(path) {
 }
 
 function isAdmin() {
-  let groups = getState(['identity', 'cognito:groups']) || [];
```
mtfranchetto, line 146: suggestion(non-blocking): it might be good extract this magic string into at least a constant so that:
- no magic strings are repeated inside the app
- if we choose to adopt properly OAuth2 on the frontend this key will be configured frontend side and thus having that as a constant is better and more easily replaceable by an ENV variable

api/PclusterApiHandler.py
```diff
@@ -533,15 +534,23 @@ def get_instance_types():
     return {"instance_types": sorted(instance_types, key=lambda x: x["InstanceType"])}
 
 
+def _get_user_roles(decoded):
+    return decoded[USER_ROLES_CLAIM] if USER_ROLES_CLAIM in decoded else ["user"]
+
+
+
 def get_identity():
```
mendaomn, line 542: issue(blocking): I'd take this chance to add our first test. It can be just an happy path test where we test that the function is returning the expected `decoded` object

frontend/src/store.js
tmscarla: Ok, going to add a `constants.ts` file and set the magic string there

api/tests/test_pcluster_api_handler.py
```diff
@@ -0,0 +1,20 @@
+import pytest
```
mtfranchetto, line 1: todo: please add to the readme how to run Python test as we'll need it to setup the test pipeline

mtfranchetto (approved):

mendaomn (approved):

PR #174, opened 2022-07-08
Convert `_app.js` to Typescript

### LINKED ISSUE #173: Use `ace-builds` to remove runtime imports of `ace`
suggestion(non-blocking): you can use `ace-builds` to remove runtime imports of `ace` (in this PR or another one), which is not strictly related to the Next but it's good to do

_Originally posted by @mtfranchetto in https://github.com/aws-samples/pcluster-manager/pull/170#discussion_r915981246_

mendaomn:
## Why

This PR converts `_app.js` to TypeScript. It also addresses #173 

## Changes

- remove `SafeHydrate` as we decided to rely on `next/dynamic` anyway
- convert the whole file to TypeScript, fixing imprecisions in the custom theme options
- use a locally hosted copy of the `ace` editor as suggested in #173 

Closes #173 


frontend/src/pages/_app.tsx
```diff
@@ -0,0 +1,108 @@
...
+      </Head>
+      <Provider store={store}>
+        <ThemeProvider theme={theme}>
+          <SnackbarProvider>
+            <Component {...pageProps} />
+          </SnackbarProvider>
+        </ThemeProvider>
+      </Provider>
+      <div id="editor"></div>
+      <Script
```
mtfranchetto, line 83: thought(non-blocking): it would have been better to directly import this thing inside the bundle directly, but we'll dive into this issue another time

PR #176, opened 2022-07-08
OIDC parametrization

eantonin:
**Description of changes:**

This PR aims to parametrize the values needed for OIDC auth. Decoupling it from Cognito.
The goal is to have everything working as before, but being able to support another Auth provider _in the future_ more easily.
The defaults are meant to make the code work as it was before this change.

This PR depends on https://github.com/aws-samples/pcluster-manager/pull/172 and should be merged after that. (Done, rebased after that)

We decided to add `openid` as default in the `scopes`, as that should be the bare minimum to make the system work, in fact, without that, we would not be able to get the `idToken`.
The data that were taken by Cognito CLI before are taken by the `idToken` at the moment. We are putting all the data in the `idToken` in the `attributes` section at the moment, a more fine choice can be made in the future.

Cookies are now secure/httpOnly/sameSite to adhere to best practices.

The `AUDIENCE` environment variable was added as it was mandatory to decode the `idToken`

The fixture was added following FLASK docs on https://flask.palletsprojects.com/en/2.1.x/testing/

**Testing**

Change was implemented testing locally. Then I deployed the new lambda and the new CF stacks manually to confirm everything was working also remotely.

After doing the changes requested by reviewers, I tested only remotely updating the Lambda.

Run `pytest` before submitting the PR to verify test were passing

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -669,18 +669,24 @@ def login():
     )
 
     access_token = code_resp.json().get("access_token")
+    id_token = code_resp.json().get("id_token")
+    refresh_token = code_resp.json().get("refresh_token")
     if not access_token:
         return redirect(AUTH_URL, code=302)
 
     # give the jwt to the client for future requests
     resp = redirect("/index.html", code=302)
-    resp.set_cookie("accessToken", access_token)
+    resp.set_cookie("accessToken", access_token, httponly=True, secure=True, samesite="lax")
+    resp.set_cookie("idToken", id_token, httponly=True, secure=True, samesite="lax")
+    resp.set_cookie("refreshToken", refresh_token, httponly=True, secure=True, samesite="lax")
     return resp
 
 
 def logout():
     resp = redirect("/login", code=302)
     resp.set_cookie("accessToken", "", expires=0)
+    resp.set_cookie("idToken", "", expires=0)
+    resp.set_cookie("refreshToken", "", expires=0)
     return resp
 
 
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -678,9 +678,9 @@ def login():
 
     # give the jwt to the client for future requests
     resp = redirect("/index.html", code=302)
-    resp.set_cookie("accessToken", access_token, httponly=True, secure=True, samesite="lax")
-    resp.set_cookie("idToken", id_token, httponly=True, secure=True, samesite="lax")
-    resp.set_cookie("refreshToken", refresh_token, httponly=True, secure=True, samesite="lax")
+    resp.set_cookie("accessToken", access_token, httponly=True, secure=True, samesite="Lax")
+    resp.set_cookie("idToken", id_token, httponly=True, secure=True, samesite="Lax")
+    resp.set_cookie("refreshToken", refresh_token, httponly=True, secure=True, samesite="Lax")
     return resp
 
 
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -45,9 +45,8 @@ try:
 except Exception:
     pass
 
+
 # Helpers
-
-
 def running_local():
     return not os.getenv("AWS_LAMBDA_FUNCTION_NAME")
 
@@ -525,7 +524,8 @@ def get_instance_types():
         ec2 = boto3.client("ec2")
     filters = [
         {"Name": "current-generation", "Values": ["true"]},
-        {"Name": "instance-type", "Values": ["c5*", "c6*", "g4*", "g5*", "hpc*", "p3*", "p4*", "t2*", "t3*", "m6*", "r*"]},
+        {"Name": "instance-type",
+         "Values": ["c5*", "c6*", "g4*", "g5*", "hpc*", "p3*", "p4*", "t2*", "t3*", "m6*", "r*"]},
     ]
     instance_paginator = ec2.get_paginator("describe_instance_types")
     instances_paginator = instance_paginator.paginate(Filters=filters)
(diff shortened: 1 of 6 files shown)


api/PclusterApiHandler.py
```diff
@@ -41,12 +48,20 @@
...
         CLIENT_SECRET = secret.get("clientSecret")
+        if not SCOPES_LIST:
+            SCOPES_LIST = "openid"
+        elif "openid" not in SCOPES_LIST:
+            SCOPES_LIST += " openid"
+        if not AUTH_URL:
+            AUTH_URL = f"{AUTH_PATH}/login?response_type=code&client_id={CLIENT_ID}&scope={SCOPES_LIST}&redirect_uri={REDIRECT_URL}"
+        if not JWKS_URL:
+            JWKS_URL = os.getenv("JWKS_URL",
+                                 f"https://cognito-idp.{REGION}.amazonaws.com/{USER_POOL_ID}/" ".well-known/jwks.json")
```
BarcoMasile, line 59: These setups should be outside of the if since this if branch only covers the case that cognito info (userpool id, client-id and client-secret) are stored in secrets manager.

Plus, the AUTH_PATH is not needed and AUTH_URL should cover the whole URL to use for user redirection, only missing the query params part (so from the '?' on)

api/PclusterApiHandler.py
```diff
@@ -631,37 +642,41 @@ def set_user_role():
...
     if not access_token:
-        return redirect(auth_redirect_path, code=302)
+        return redirect(AUTH_URL, code=302)
 
     # give the jwt to the client for future requests
     resp = redirect("/index.html", code=302)
-    resp.set_cookie("accessToken", access_token)
+    resp.set_cookie("accessToken", access_token, httponly=True, secure=True, samesite="lax")
+    resp.set_cookie("idToken", id_token, httponly=True, secure=True, samesite="lax")
+    resp.set_cookie("refreshToken", refresh_token, httponly=True, secure=True, samesite="lax")
```
BarcoMasile, line 671: Looks like "lax" should in fact be "Lax"

api/PclusterApiHandler.py
```diff
@@ -545,16 +557,15 @@ def get_identity():
...
-        decoded = jwt_decode(access_token, USER_POOL_ID)
+        decoded = jwt_decode(access_token)
+        decoded_id = jwt_decode(id_token, audience=AUDIENCE, access_token=access_token)
         username = decoded.get("username")
         if username:
-            cognito = boto3.client("cognito-idp")
-            filter_ = f'username = "{username}"'
-            user = cognito.list_users(UserPoolId=USER_POOL_ID, Filter=filter_)["Users"][0]
-            decoded["attributes"] = {ua["Name"]: ua["Value"] for ua in user["Attributes"]}
+            decoded["attributes"] = {key: value for key, value in decoded_id.items()}
```
BarcoMasile, line 568: Here we don't need the check for the username, since this is related to the previous implementation based strongly on Cognito.
We could just have what is now on line 568 withouth the retrieval of the username and without the `if username`.

api/PclusterApiHandler.py
```diff
@@ -33,6 +33,13 @@
 SECRET_ID = os.getenv("SECRET_ID")
 ENABLE_MFA = os.getenv("ENABLE_MFA")
 SITE_URL = os.getenv("SITE_URL", API_BASE_URL)
+SCOPES_LIST = os.getenv("SCOPES_LIST")
+REGION = os.getenv("AWS_DEFAULT_REGION")
+REDIRECT_URL = os.getenv("REDIRECT_URL", f"{SITE_URL}/login")
```
BarcoMasile, line 38: REDIRECT_URL could just be `f"{SITE_URL}/login"` without the ENV var since we know where to redirect to

api/PclusterApiHandler.py
```diff
@@ -641,37 +652,41 @@ def set_user_role():
 
 
 def login():
```
mendaomn, line 654: issue(blocking): We should add a test for this function, as we started doing with https://github.com/aws-samples/pcluster-manager/pull/172

We could go for an easy first test for this function, for instance:

```
given a request with code
  when the auth provider responds successfully
  it should add the accessToken cookie to the response
  it should add the idToken cookie to the response
  it should add the refreshToken cookie to the response
```

api/PclusterApiHandler.py
```diff
@@ -33,7 +33,14 @@
...
 ENABLE_MFA = os.getenv("ENABLE_MFA")
 SITE_URL = os.getenv("SITE_URL", API_BASE_URL)
+SCOPES_LIST = os.getenv("SCOPES_LIST")
+REGION = os.getenv("AWS_DEFAULT_REGION")
+TOKEN_URL = os.getenv("TOKEN_URL", f"{AUTH_PATH}/oauth2/token")
+AUTH_URL = os.getenv("AUTH_URL", f"{AUTH_PATH}/login")
+JWKS_URL = os.getenv("JWKS_URL")
+AUDIENCE = os.getenv("AUDIENCE")
 USER_ROLES_CLAIM = os.getenv("USER_ROLES_CLAIM", "cognito:groups")
+REDIRECT_URL = f"{SITE_URL}/login"
```
mendaomn, line 43: suggestion(non-blocking): Shall we document this new stuff? (the README maybe?)

api/PclusterApiHandler.py
eantonin: sure, will do ASAP

api/PclusterApiHandler.py
mendaomn: I see. Would this affect our ability to run the backend locally though?

eantonin: No, I don't think so, as we already have a runbook with the env variables needed to run locally

BarcoMasile (approved):

PR #179, opened 2022-07-09
Cluster Templates

sean-smith:
## Description

* Pre-Defined Cluster templates for common applications
* All templates are hosted in `/templates` directory on Github

## Demo

https://user-images.githubusercontent.com/5545980/178086496-5946f18d-d99d-4f3f-96a9-263c76815dbc.mov

## Testing

Change the path of the Github URL in [api/PclusterApiHandler.py#210](https://github.com/aws-samples/pcluster-manager/pull/179/files#diff-5424a9a7f7cf5ddc8d3e420f077ac509a3b37f81c8cbf7c9bac5f8695fc9a8e2)

**from**
```bash
https://raw.githubusercontent.com/aws-samples/pcluster-manager/main/{template_path}
```

**to**
```bash
https://raw.githubusercontent.com/sean-smith/pcluster-manager/template/{template_path}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ infrastructure/cognitolambda/node_modules
 .DS_Store
 .hugo_build.lock
 public/*
+package.json
+package-lock.json
--- a/app.py
+++ b/app.py
@@ -27,6 +27,7 @@ from api.PclusterApiHandler import (
     ec2_action,
     get_aws_config,
     get_cluster_config,
+    get_cluster_template,
     get_custom_image_config,
     get_dcv_session,
     get_identity,
@@ -93,6 +94,11 @@ def run():
     @authenticated()
     def get_cluster_config_():
         return get_cluster_config()
+
+    @app.route("/manager/get_cluster_template")
+    @authenticated()
+    def get_cluster_template_():
+        return get_cluster_template()
 
     @app.route("/manager/get_custom_image_configuration")
     @authenticated()
--- a/frontend/src/old-pages/Configure/Source.js
+++ b/frontend/src/old-pages/Configure/Source.js
@@ -14,7 +14,7 @@ import React from 'react';
 import { useState, setState, getState, clearState } from '../../store'
 import { loadTemplate } from './util'
 import { findFirst } from '../../util'
-import { GetConfiguration } from '../../model'
+import { GetConfiguration, GetClusterTemplate } from '../../model'
 
 import jsyaml from 'js-yaml';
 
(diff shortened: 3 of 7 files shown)


.gitignore
```diff
@@ -7,3 +7,5 @@ infrastructure/cognitolambda/node_modules
 .DS_Store
 .hugo_build.lock
 public/*
+package.json
+package-lock.json
```
mendaomn, line 11: issue(blocking): Careful, these two cannot be gitignored :)

app.py
```diff
@@ -94,6 +95,11 @@ def ec2_action_():
     def get_cluster_config_():
         return get_cluster_config()
 
+    @app.route("/manager/get_cluster_template")
```
mtfranchetto, line 98: suggestion: cluster templates are plain strings and can probably be inlined in the JS bundle as they add few extra bytes (nothing comparable to third party modules). What I suggest is to remove this extra configuration management and use something like [next-plugin-yaml](https://www.npmjs.com/package/next-plugin-yaml) to inline the templates frontend side. This has several benefits:
- no new route backend side
- no new custom requests frontend side and error management
- no manual transform of yaml -> js: this gets handled automatically by `next-plugin-yaml`
- simplified testing: no need to manually edit URL to fetch template, simply edit the files and see the result in the browser

frontend/src/old-pages/Configure/Source.js
```diff
@@ -181,7 +194,13 @@ function Source() {
               {
                 value: "wizard",
                   label: "Wizard",
-                  description: "Choose this to start a new cluster configuration."
+                  description: <Box margin={{bottom: "xs"}} >
```
mtfranchetto, line 197: Screenshot 2022-07-11 at 11 39 18
issue: I personally find very confusing this kind of UX where I can pick a template from the list but I also have another option called in the same way (Template) beneath it. While this is not a major issue it still needs some UX attention, and can be reworked in different ways:
- rename the Template section: maybe with a less fancy name like "Wizard with existing configuration"?
- rename the new templates feature: presets?
- add another option for our users: Empty wizard / Wizard with preset / Wizard from existing configuration / and so on

@sean-smith what do you think?

mtfranchetto (changes requested):

app.py
sean-smith: I like this approach! I think the only downside is we can't update between releases then right?

frontend/src/old-pages/Configure/Source.js
sean-smith: I think having two Wizard sections might be confusing but I could re-work the template section to be either "Start from an existing template" or "upload your own template" maybe with a toggle UI element, or a setting in the list "Upload a default template". Definitely want some PM eyes on this feature.

app.py
mtfranchetto: Yes, correct, but we can start small by making those files static and then, if we have to for some reasons, write some code to add dynamic imports from cloud. Anyway, I think that if we invest in testing, build times and automated deployments it is going to be much easier to release something thus static files should be ok imho.

PR #184, opened 2022-07-12
Setup GitHub Actions to run tests on opened, reopened or merged PR

BarcoMasile:
### Description
This PR introduces one workflow definition to run frontend tests and backend tests in parallel, every time a PR is opened, reopened or merged successfully

### Issues
[Asana task](https://app.asana.com/0/1202530481823466/1202523788825836/f)


.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,53 @@
+
+name: Pull Request
+
+on:
+  pull_request:
+    types:
+      - opened
```
mtfranchetto, line 7: question(possibly-blocking): does this catch the case when someone does `git push` again on a branch? I guess that `synchronized` should be added [link](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)

mtfranchetto: From Github:

> By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened.

we can get rid of `types` entirely

BarcoMasile: actually now that I pushed a second revision, the check didn't run, so it's exactly as you said.
I will remove it and add a comment. Thank you

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,51 @@
...
+jobs:
+  frontend-tests:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - uses: actions/setup-node@v3
+        name: Setup Node version 16q
```
mtfranchetto, line 20: nitpick" I'll remove the version entirely, leaving only "Setup Node version", since the version is configured inside .nvmrc and can go out of sync, causing confusion

BarcoMasile: Absolutely right, it could cause more than a headache in the future.
Pushing new rev with this change. Thanks!

PR #190, opened 2022-07-14
Make wizard compliant with limits for external storage attachments

ermanno:
### Notes
This patch makes the wizard compliant with the limits for external
storage attachments.

When creating a new storage, we check that we are not violating the
constraint of the maximum number of storage of that given type we can
attach. The toggle that allows to switch between creation and attachment
of an existing storage is disabled according to the constraints of the
maximum number of storage that can be created or of existing storage that
can be attached.

### Tests
- Unit tests.
- Manual tests:
  - Attached Efs storage (creation).
  - Attached 20 Efs storages (existing), verifying that the toggle is automatically set to "use existing" and disabled (greyed out).
  - When the 20th existing Efs storage is attached, the toggle of the Efs storage to create is also disabled.

### DIFF
--- a/frontend/src/__tests__/storagCreationValidation.test.tsx
+++ b/frontend/src/__tests__/storagCreationValidation.test.tsx
@@ -1,0 +1,87 @@
+import { canCreateStorage, canAttachExistingStorage } from "../old-pages/Configure/Storage";
+
+
+describe("Given a function to determine whether we can create a storage of a given type", () => {
+    describe("when the storage type is not available", () => {
+        it("should not allow the creation of a new storage", () => {
+            expect(canCreateStorage(null, [], {})).toBeFalsy();
+        });
+    });
+    describe("when the storage type does not support creation", () => {
+        it("should not allow the creation of a new storage", () => {
+            expect(canCreateStorage("FsxOntap", [], {})).toBeFalsy();
+        });
+    });
+    describe("when the attached storages are not available", () => {
+        it("should allow the creation of a new storage", () => {
+            expect(canCreateStorage("Efs", null, {})).toBeTruthy();
+        });
+    });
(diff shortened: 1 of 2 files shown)


frontend/src/old-pages/Configure/Storage.js
```diff
@@ -494,6 +504,7 @@ function StorageInstance({index}) {
   const existingPath = STORAGE_TYPE_PROPS[storageType].mountFilesystem ? [...settingsPath, 'FileSystemId'] : [...settingsPath, 'VolumeId'];
```
sean-smith, line 504: This line causes an exception when the "Add Storage" button is clicked but no storage type is selected

ermanno: Yes, I had a discussion about this with @mtfranchetto, but he was telling me that this case shouldn't happen in this component (the line above with the `"none"` default shouldn't be needed).

ermanno: This was my understanding.

frontend/src/__tests__/storagCreationValidation.test.tsx
```diff
@@ -0,0 +1,87 @@
+import { canCreate, canAttachExisting } from "../old-pages/Configure/Storage";
+
+
+describe("Given a function to determine whether we can create a storage of a given type", () => {
+    describe("when the storage type is null", () => {
+        it("should return false", () => {
```
mtfranchetto, line 6: suggestion: it's better to avoid direct references to the implementation in the test spec to do a proper decoupling.
In this case "should return false" may be rephrased as "should not allow the creation of a new storage". Same goes for "when the storage type is null". 

Overall great tests!

mtfranchetto (approved):

PR #192, opened 2022-07-18
Realign PC API template reflecting aws-parallelcluster changes

ermanno:
### Notes

This patch realigns the PC API template, reflecting the changes in:

https://github.com/aws/aws-parallelcluster/blob/develop/api/infrastructure/parallelcluster-api.yaml

among those, the missing policies that were added in:

https://github.com/aws/aws-parallelcluster/pull/4194

### DIFF
--- a/infrastructure/parallelcluster-api.yaml
+++ b/infrastructure/parallelcluster-api.yaml
@@ -612,6 +612,7 @@ Resources:
                   - s3.data-source.lustre.fsx.amazonaws.com
           - Action:
               - lambda:CreateFunction
+              - lambda:TagResource
               - lambda:DeleteFunction
               - lambda:GetFunctionConfiguration
               - lambda:GetFunction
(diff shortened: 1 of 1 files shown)


sean-smith:
This file should be unnecessary now that https://github.com/aws/aws-parallelcluster/pull/3886 has been released in `3.1.3`. Suggest removing this file altogether and just referencing the stack on S3 directly.

ermanno:
Talking with @sean-smith it turns out that we don't need this template in the repo, we can use the PI API template hosted on S3. Closing this PR.

charlesg3:
My question is if we need to make corresponding changes in our own infrastructure to reflect the changes made in the PC infrastructure? We use the same imagebuilder process.

ermanno:
Regarding this, afaik (correct me if I am wrong), we reference the template with `TemplateURL: !Sub https://pcluster-manager-${AWS::Region}.s3.${AWS::Region}.amazonaws.com/parallelcluster-api.yaml` so we should "get" the changes made to that template automatically when we deploy the API.

PR #193, opened 2022-07-19
Add option, in the wizard, to enable memory based scheduling in Slurm

maurizp:
When creating a new cluster, we allow the customer to use the
ConstrainRAMSpace option available in Slurm during the job
submission.

## Description

We want to meet features parity with PCluster 3.2.0 and it includes the support of memory based scheduling

## Changes

* Added a toggle to enable memory based scheduling

## How Has This Been Tested?

* Deployed a new stack with PCluster 3.2.0b2 and tested
* Executed a DryRun with option enabled

Screenshot 2022-07-19 at 07 52 53

* Executed a DryRun with option disabled (no new fields in the yaml)
* Executed a DryRun with option enabled and then disabled the toggle (no new fields in the yaml)

## References

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #199, opened 2022-07-20
Add schedulable memory parameter in the queue section of the Wizard

eantonin:
## Description

Add schedulable memory parameter in the queue section of the Wizard to reach feature parity with pcluster 3.2.0

(The flag to enable this param only with 3.2.0 will be added in a subsequent CR in addition to other related parameters)

## Changes

1. Make configurable the HelpTextInput component
2. Add schedulable memory parameter
3. Add localization of strings for the interested part of code

## How Has This Been Tested?

Tested locally connecting to a PC version using 3.2.0b2, see video below

https://user-images.githubusercontent.com/83351137/179985446-761d5677-99c1-4814-98d9-978a3a3a1b84.mp4

## References

This was dependent of the [PR related to the memory based scheduling](https://github.com/aws-samples/pcluster-manager/pull/193)

### DIFF
--- a/frontend/locales/en/strings.json
+++ b/frontend/locales/en/strings.json
@@ -5,11 +5,61 @@
     }
   },
   "wizard": {
-     "headNode": {
-       "memoryBasedSchedulingEnabled": {
-         "label": "Slurm Memory Based Scheduling Enabled",
-         "help": "If enabled, users may use <i>--mem</i> to specify the amount of memory per node required by a job. For more information, see the <0>Slurm documentation for ConstrainRAMSpace</0>"
-       }
+    "headNode": {
+      "memoryBasedSchedulingEnabled": {
+        "label": "Slurm Memory Based Scheduling Enabled",
+        "help": "If enabled, users may use <i>--mem</i> to specify the amount of memory per node required by a job. For more information, see the <0>Slurm documentation for ConstrainRAMSpace</0>"
+     }
+   },
+    "queues": {
+      "schedulableMemory": {
+        "name": "Schedulable Memory (MiB)",
+        "description": "Amount of memory in MiB to be made available to jobs on the compute nodes of the compute resource",
+        "placeholder": "Leave empty for default",
+        "help": "The default value is 95 % of the memory advertised by EC2."
+      }
+   },
+    "multiuser": {
+      "domainName": {
+        "name": "Domain Name*",
+        "description": "The Active Directory (AD) domain that you use for identity information.",
+        "help": "This property corresponds to the sssd-ldap parameter that's called ldap_search_base."
+      },
+      "domainAddress": {
+        "name": "Domain Address*",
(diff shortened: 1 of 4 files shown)


frontend/src/old-pages/Configure/Components.tsx
```diff
@@ -636,4 +637,50 @@ function IamPoliciesEditor({
...
+      label={name}
+      errorText={error}
+      description={description}>
+    <div style={{display: "flex", flexDirection: "row", alignItems: "center", justifyContent: "space-between"}}>
+      <div style={{flexGrow: 1}}>
+        <Input
+            placeholder={placeholder}
+            value={value}
+            type={type}
+            onChange={({detail}) => {setterFunction([...path, configKey], detail.value); validationFunction();}} />
```
mendaomn, line 679: suggestion(non-blocking): for higher extensibility, you could expose an `onChange` prop to users of the component

You can think of this component as glorified `Input` field with the help tooltip, it doesn't need to implement any logic

Users of the component could decide to apply a validation function or not, for instance

eantonin: This makes perfect sense, will do

eantonin:
> Shouldn't we put this in the "Advanced Options" section as most users will want the default memory (i.e. 95% of the Total memory).

Yeah Sean, that make perfect sense, will fix

eantonin:
@[sean-smith](https://github.com/sean-smith), actually that cannot be made easily, as SchedulableMemory parameter should be a queue specific parameter, while advanced options are now compute-specific and not queue-specific, see images below

queue+advanced-options

yaml-queue-advanced

So with the current configuration I cannot put SchedulableMemory in advanced option

A possibility could be add a queue-specific advanced option section, personally I don't like it a lot, and by the way I'll consider it out of scope for this PR

frontend/locales/en/strings.json
```diff
@@ -5,11 +5,61 @@
...
+        "help": "This property corresponds to the sssd-ldap parameter that's called ldap_search_base."
+      },
+      "domainAddress": {
+        "name": "Domain Address*",
+        "description": "The URI or URIs that point to the AD domain controller that's used as the LDAP server.",
+        "help": "The URI corresponds to the sssd-ldap parameter that's called ldap_uri. The value can be a comma separated string of URIs. To use LDAP, you must add ldap:// to the beginning of the each URI."
+      },
+      "passwordSecretArn" : {
+        "name": "Password Secret ARN*",
+        "description": "The URI or URIs that point to the AD domain controller that's used as the LDAP server.",
```
eantonin, line 35: The description and help here are wrong because of a copy-paste error, will fix in a subsequent CR

mendaomn (approved):

eantonin:
Rebasing branch without further test as I'm rebasing on https://github.com/aws-samples/pcluster-manager/pull/204 which is regarding just fixes to a script

PR #200, opened 2022-07-20
Allow to specify reqired memory when launching a job

ermanno:
### Notes
This patch allows the user to specify teh required memory when launching a job, if it is run on a Slurm cluster with memory based scheduling enabled.

### Tests
- Tested with a cluster with memory based scheduling enabled, observed that the correct parameter `--mem <specified value>` was present in the request to `sbatch`.
- Tested with a cluster with memory based scheduling, without specifying the value, verified that it was not present in the `sbatch` parameters.
- Tested with a cluster without memory based scheduling, verified that the part of the dialogue that allows to specify the required memory for the job is not show.

### DIFF
--- a/frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
+++ b/frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
@@ -9,6 +9,7 @@
 // OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 // limitations under the License.
 import * as React from 'react';
+import { useTranslation } from 'react-i18next';
 import { useState, setState, getState, clearState } from '../../store'
 import { findFirst, clusterDefaultUser } from '../../util'
 import { SubmitJob, PriceEstimate } from '../../model'
@@ -147,9 +148,11 @@ function JobCostEstimate() {
 export default function JobSubmitDialog({
   submitCallback
 }: any) {
+  const { t } = useTranslation();
   const open = useState([...submitPath, 'dialog']);
   const error = useState([...submitPath, 'error']);
   const jobPath = [...submitPath, 'job'];
+  const clusterName = getState(['app', 'clusters', 'selected']);
 
   const job = useState(jobPath);
   const submitting = useState([...submitPath, 'pending']);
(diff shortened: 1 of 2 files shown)


frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
```diff
@@ -263,6 +270,23 @@ export default function JobSubmitDialog({
           </FormField>
         </SpaceBetween>
 
+        {
+          isMemBasedSchedulingEnabled && <SpaceBetween direction="vertical" size="xxs" key="mem">
+          <Header variant="h2"
+            description={t("JobSubmitDialog.requiredMemory.description")}>
+            <Trans i18nKey="JobSubmitDialog.requiredMemory.header" />
```
maurizp, line 277: Why are you using _Trans_ instead of the standard _useTranslation_ function used above in this case?

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
```diff
@@ -9,6 +9,7 @@
 // OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 // limitations under the License.
 import * as React from 'react';
+import {Trans, useTranslation} from 'react-i18next';
```
maurizp, line 12: Is _Trans_ strictly needed?

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
ermanno: That's a good point, I had started using `Trans` and then used the function because I was unable to use `Trans` for the description, trying with the function directly.

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
ermanno: Maybe not, I'll do a test using only `useTranslation` and eventually send a new revision.

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
ermanno: Okay, verified that we can do everything with the function, changed in rev2.

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
ermanno: Verified that we can do everything with the function, changed in rev2.

maurizp (approved):

PR #201, opened 2022-07-20
[scripts/downloader] Add a downloader script for syncing data.

charlesg3:
## Description

Adds a new multi-runner script that allows for the downloading of data at cluster creation. This allows for better bootstrapping of clusters that might need some small initial data that can be downloaded, thus allowing customers to continue to use PCM assuming said data is there.

## Changes

- Add a script that takes a destination an multiple source arguments and downloads each of those arguments to the destination
- Add said script to the multi-runner configuration

## How Has This Been Tested?

- Manually by creating a cluster with the multi-runner script enabled and validating that the files were successfully downloaded.

By creating a cluster with the following extension which specifies a destination and two arguments to download from:

image

I can login to the cluster and see that `index.html` and `parallelcluster-api.yaml` have been downloaded at cluster start time.

```
$ pcluster ssh -n downloader4 -i ~/keys/enguard.pem
The authenticity of host '3.12.197.2 (3.12.197.2)' can't be established.
ECDSA key fingerprint is SHA256:6M5OcTKE75IODge/Ld77VIgGMlyPVUuqJOsNWGTtYRU.
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added '3.12.197.2' (ECDSA) to the list of known hosts.
Last login: Wed Jul 20 22:07:39 2022

       __|  __|_  )
       _|  (     /   Amazon Linux 2 AMI
      ___|\___|___|

https://aws.amazon.com/amazon-linux-2/
[ec2-user@ip-10-0-0-77 ~]$ ls /tmp
cfn-hup-stderr---supervisor-nde8kxfw.log  cfn-hup-stdout---supervisor-l_mszwrw.log  dna.json  extra.json  hsperfdata_root  index.html  parallelcluster-api.yaml  systemd-private-48ef0313e2c5454887663506cc30f718-chronyd.service-lCu4MJ  systemd-private-48ef0313e2c5454887663506cc30f718-cups.service-ysGmmW  wait_condition_handle.txt
[ec2-user@ip-10-0-0-77 ~]$ 
```


eantonin:
@charlesg3 

I have two questions:

1. Have you tested with URLs starting with s3 and http too?
2. Are these scripts documented somewhere to help the customer understand how they work?

resources/scripts/downloader.sh
```diff
@@ -0,0 +1,47 @@
...
+# Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+# with the License. A copy of the License is located at
+#
+# http://aws.amazon.com/apache2.0/
+#
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
```
eantonin, line 13: issue(non-blocking): This is often considered a bad practice, see [this Stack Overflow post](https://stackoverflow.com/questions/19622198/what-does-set-e-mean-in-a-bash-script)

resources/scripts/downloader.sh
```diff
@@ -0,0 +1,47 @@
...
+
+while [[ $# -gt 0 ]]
+do
+SOURCE="$1"
+echo "Downloading ${SOURCE} to ${DEST}"
+
+if [[ $SOURCE =~ s3:// || $SOURCE =~ S3:// ]]; then
+  aws s3 cp $SOURCE .
+elif [[ $SOURCE =~ http:// || $SOURCE =~ HTTP:// || $SOURCE =~ https:// || $SOURCE =~ HTTPS:// ]]; then
+  wget -q ${SOURCE}
```
eantonin, line 40: suggestion(non-blocking): You could lower case $SOURCE in the condition and avoid double checks

resources/scripts/downloader.sh
eantonin, line 13: issue(non-blocking): This is often considered a bad practice, see [this Stack Overflow post](https://stackoverflow.com/questions/19622198/what-does-set-e-mean-in-a-bash-script)

resources/scripts/downloader.sh
eantonin, line 40: suggestion(non-blocking): You could lower case $SOURCE in the conditions and avoid double checks

resources/scripts/downloader.sh
```diff
@@ -0,0 +1,47 @@
...
+# or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+# OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+
+DEST=$1
+
+if [ "$DEST" == "" ]; then
+  echo "You must specify a destination directory."
```
eantonin, line 18: question(non-blocking): Are this echos readable to someone?

eantonin (approved):
Approving as there are no-blocking comments

charlesg3:
> @charlesg3
> 
> I have two questions:
> 
> 1. Have you tested with URLs starting with http too?
> 2. Are these scripts documented somewhere to help the customer understand how they work?
> 
> Second question is kind of important to me because seeing the image above, it was not really clear to me what the script was trying to do, Source and Arg3 are kind of misleading, someone could think Arg3 concatenates with Source for example.
> 
> If we do not have a simple way to make those parameter more understandable via code, documentation could help. We could start documenting this script, and take that as a best practice for future scripts. What do you think?

Where do you recommend putting this documentation? I am happy to provide this wherever makes sense.

resources/scripts/downloader.sh
charlesg3: I believe they should be in the cloudwatch logs

resources/scripts/downloader.sh
charlesg3: done.

resources/scripts/downloader.sh
charlesg3: done.

eantonin:
> > @charlesg3
> > I have two questions:
> > 
> > 1. Have you tested with URLs starting with http too?
> > 2. Are these scripts documented somewhere to help the customer understand how they work?
> > 
> > Second question is kind of important to me because seeing the image above, it was not really clear to me what the script was trying to do, Source and Arg3 are kind of misleading, someone could think Arg3 concatenates with Source for example.
> > If we do not have a simple way to make those parameter more understandable via code, documentation could help. We could start documenting this script, and take that as a best practice for future scripts. What do you think?
> 
> Where do you recommend putting this documentation? I am happy to provide this wherever makes sense.

@charlesg3 Maybe the area in which the documentation can fit better is a https://pcluster.cloud/02-tutorials.html? Do you agree?

resources/scripts/downloader.sh
```diff
@@ -0,0 +1,50 @@
...
+  exit 1
+fi
+shift
+
+mkdir -p ${DEST}
+pushd ${DEST}
+
+if [[ $# -eq 0 ]]; then
+  echo "You must specify at leats one file to download."
```
eantonin, line 31: (nitpick, non blocking) Typo leats -> least

eantonin (approved):
approving with nitpick

resources/scripts/downloader.sh
charlesg3: Fixed.

eantonin (approved):
LGTM

PR #202, opened 2022-07-20
Cost explorer

AndyHoNYC:
## Description

This adds a tab called "Cost" which summarizes Cluster cost by pulling in data from cost-explorer and graphing it by instance type. Also, this includes a button to activate cost allocation tags, which is a necessary pre-requisite for cost explorer.

## Changes

- Included new Cost.tsx file
- Added route `/manager/graph_data?cluster_name=Cluster&Start=Start&End=End/`
- Added route `/manager/activate_tags/`
- Added route `/manager/check_tag_status/`

## How Has This Been Tested?

- Tested locally using created clusters

https://user-images.githubusercontent.com/68314782/180088672-fad78396-854a-425f-92c8-7c8152289d64.mov

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -41,6 +41,15 @@ JWKS_URL = os.getenv("JWKS_URL")
 AUDIENCE = os.getenv("AUDIENCE")
 USER_ROLES_CLAIM = os.getenv("USER_ROLES_CLAIM", "cognito:groups")
 REDIRECT_URL = f"{SITE_URL}/login"
+PCluster_tags = ['parallelcluster:attributes', 
+        'parallelcluster:cluster-name', 'parallelcluster:compute-resource-name', 'parallelcluster:filesystem', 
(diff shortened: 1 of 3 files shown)


api/PclusterApiHandler.py
```diff
@@ -695,9 +701,50 @@ def _get_params(_request):
...
+        return {"message": str(e)}, 500
 
+# Retrieve usage data from Cost Explorer API
+def graph_data():
+    try:
+        cost_explorer = boto3.client('ce')
+        data = cost_explorer.get_cost_and_usage(
+            TimePeriod={
+                'Start': '2022-07-06',
+                'End': '2022-07-12'
```
BarcoMasile, line 731: Question: I see these hardcoded values here, is it ok? Since no parameters are taken from the API request, this seems like it always returns values for the time range 2022-07-06 to 2022-07-12, instead of the one selected from the UI

sean-smith: This isn't the right commit. Hang on with reviewing, new commit makes this dynamic i.e. today - 7

AndyHoNYC:
Closing in favor of #162

PR #203, opened 2022-07-21
Fixing violation related to Defend Against Reverse Tabnabbing Attack

maurizp:
Reverse tabnabbing is an attack that is possible when a link is
configured to open a new tab in a browser. If done incorrectly,
the new page is able to control the referrer and opener
objects of the parent window, and use that control to replace
the parent site with a phishing site.
The mitigation is add the noopener and noreferrer values in
 the rel attribute in <a> tags.

## Description

Reverse tabnabbing is an attack that is possible when a link is
configured to open a new tab in a browser. If done incorrectly,
the new page is able to control the referrer and opener
objects of the parent window, and use that control to replace
the parent site with a phishing site.
The mitigation is add the noopener and noreferrer values in
 the rel attribute in <a> tags.

## Changes

Added the *noopener* tag

## How Has This Been Tested?

Locally tested. The new tag is added.

## References


mendaomn (approved):

mendaomn:
LGTM, let's merge it. AFAIK, this type of rules should be already included in the linter, I'll look into it

PR #209, opened 2022-07-22
`[Configure/Source]` Validate the cluster's name regex

charlesg3:
## Description

Validate the cluster name regex in the Source tab.

## Changes

Use the same regular expression that will be used in cluster creation in the source tab so that the error is fixed earlier.

## How Has This Been Tested?

This has been tested manually.
image

### DIFF
--- a/frontend/src/old-pages/Configure/Source.tsx
+++ b/frontend/src/old-pages/Configure/Source.tsx
@@ -68,6 +68,9 @@ function sourceValidate(suppressUpload = false) {
   } else if(clusterNames.has(clusterName)) {
     setState([...sourceErrorsPath, 'clusterName'], `Cluster with name ${clusterName} already exists. Please choose a unique name.`);
     valid = false;
+  } else if(!/^[a-zA-Z][a-zA-Z0-9-]+$/.test(clusterName)) {
+    setState([...sourceErrorsPath, 'clusterName'], `Cluster name ${clusterName} doesn't match the format /^[a-zA-Z][a-zA-Z0-9-]+$/. Please choose a name consisting of only upper and lower-case letters, numbers and dashes.`);
+    valid = false;
   } else {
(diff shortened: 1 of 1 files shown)


frontend/src/old-pages/Configure/Source.tsx
```diff
@@ -68,6 +68,9 @@ function sourceValidate(suppressUpload = false) {
   } else if(clusterNames.has(clusterName)) {
     setState([...sourceErrorsPath, 'clusterName'], `Cluster with name ${clusterName} already exists. Please choose a unique name.`);
     valid = false;
+  } else if(!/^[a-zA-Z][a-zA-Z0-9-]+$/.test(clusterName)) {
+    setState([...sourceErrorsPath, 'clusterName'], `Cluster name ${clusterName} doesn't match the format /^[a-zA-Z][a-zA-Z0-9-]+$/. Please choose a name consisting of only upper and lower-case letters, numbers and dashes.`);
```
mtfranchetto, line 72: chore: as we introduced `react-i18next` it would be good to avoid hardcoded labels, [this a PR](https://github.com/aws-samples/pcluster-manager/pull/199/files#diff-0fd86d36406df3d671f91ce4bd03cc5c0240569f2849736ab149c911d40998c4) to show the process

apart from this, the change is good!

charlesg3: Since this is outside of a component I used `i18next` directly to satisfy this. LMK if there's issues with that (otherwise, if not then I'll do a pass on converting a lot more of our strings to follow this pattern)

mtfranchetto: no issues at all!

charlesg3: related: https://github.com/aws-samples/pcluster-manager/pull/218

PR #215, opened 2022-07-23
`[components/ValidationErrors]` Remove MUI.

charlesg3:
## Description

Remove the use of MUI in the ValidationErrors component.

## How Has This Been Tested?

Looking at the 3 options for success, warning and error for a cluster config:

image

image

## References

### DIFF
--- a/frontend/locales/en/strings.json
+++ b/frontend/locales/en/strings.json
@@ -335,6 +335,14 @@
       }
     }
   },
+  "components": {
+    "ValidationErrors": {
+      "update": "Update",
+      "validation": "Validation",
+      "errors": "Errors",
+      "warnings": "Warnings"
+    }
+  },
   "JobSubmitDialog": {
     "requiredMemory": {
       "header": "Required memory",
--- a/frontend/src/components/ValidationErrors.tsx
+++ b/frontend/src/components/ValidationErrors.tsx
@@ -10,57 +10,45 @@
 // limitations under the License.
 
 // UI Elements
-import { useTheme } from '@mui/material/styles';
+import { Icon } from "@awsui/components-react";
 
-import List from '@mui/material/List';
-import ListItem from '@mui/material/ListItem';
-import ListItemText from '@mui/material/ListItemText';
-
-// Icons
-import CheckIcon from '@mui/icons-material/Check';
-import CloseIcon from '@mui/icons-material/Close';
-
-export default function ValidationErrors(props: any) {
-  const theme = useTheme();
+export default function ValidationErrors({errors}: any) {
   const colorMap = (level: any) => {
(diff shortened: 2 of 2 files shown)


frontend/src/components/ValidationErrors.tsx
```diff
@@ -10,57 +10,45 @@
...
-
-// Icons
-import CheckIcon from '@mui/icons-material/Check';
-import CloseIcon from '@mui/icons-material/Close';
-
-export default function ValidationErrors(props: any) {
-  const theme = useTheme();
+export default function ValidationErrors({errors}: any) {
   const colorMap = (level: any) => {
     // @ts-expect-error TS(7053) FIXME: Element implicitly has an 'any' type because expre... Remove this comment to see the full error message
```
mendaomn, line 17: It's not an easy one, but we could try and fix this by annotating `(level: string)`

The real fix would be typing the `errors` but that would involve changing how we store errors in the store, so it may not be feasible

frontend/src/components/ValidationErrors.tsx
```diff
@@ -10,57 +10,45 @@
...
     <div>
-      {colored(props.errors.message, success)}
+      {colored(errors.message, success)}
       {configErrors &&
         <div className="validation-errors">
-          Validation {props.errors.configurationValidationErrors ? "Errors" : "Warnings"}:
-          <List>
-            {configErrors.map((error: any, i: any) => <ListItem disablePadding key={i}><ListItemText primaryTypographyProps={{ style: {color: colorMap(error.level)} }} primary={`${error.type}: ${error.message}`} /></ListItem>)}
-          </List>
+          Validation {errors.configurationValidationErrors ? "Errors" : "Warnings"}:
```
mendaomn, line 44: Shall we internationalize this part as well? I understand this was not the point of the PR though, so I'll leave it up to you to decide

frontend/src/components/ValidationErrors.tsx
```diff
@@ -10,57 +10,45 @@
...
+          {configErrors.map((error: any, i: any) => <div style={{color: colorMap(error.level)}} key={i}>{`${error.type}: ${error.message}`}</div>)}
         </div>
       }
       {updateErrors &&
         <div className="validation-errors">
           Update Errors:
```
mendaomn, line 50: Same as above about i18n

frontend/locales/en/strings.json
```diff
@@ -335,6 +335,14 @@
       }
     }
   },
+  "componets": {
```
mendaomn, line 338: issue(blocking): small typo on `components`, this would break the UI

mendaomn (approved):

PR #216, opened 2022-07-25
Return 4xx in case of unauthenticated/unauthorized requests

ermanno:
### Notes
This patch makes the API return:
- 400 if the login request does not contain a `code` (in the case of the OIDC provider redirecting the user after a successful login).
- 401 if the access token in the login or authenticate calls is missing.
- 401 if during authentication there is an issue with the signature.
- 403 if during authentication the group is not guest and is not among the groups in the decoded JWT.

### Tests
- Unit tests
- Manual tests with Postman:
  - Invoke a protected API without authentication, expecting 401.
  - Through Postman, get a JWT through Cognito and invoke the same API attaching the JWT of the authentication, and get a 200.
  - Modify a character in the JWT in the central part of the token to generate an invalid signature, expecting 401.
  - Create a user on Cognito that is not assigned to any group, generate a JWT and then expect a 403.

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -138,24 +138,24 @@ def authenticate(group):
 
     access_token = request.cookies.get("accessToken")
     if not access_token:
-        return auth_redirect()
+        return abort(401)
+
     try:
         decoded = jwt_decode(access_token)
     except jwt.ExpiredSignatureError:
-        return auth_redirect()
-    except jose.exceptions.JWSSignatureError:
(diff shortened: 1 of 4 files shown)


BarcoMasile:
about this line

> 400 if the login request does not contain a code

I would add between parenthesis (in the case of the OIDC provider redirecting the user after a successful login)

BarcoMasile (dismissed):

BarcoMasile (approved):

PR #220, opened 2022-07-26
Redirect to authorization server upon 401/403 status codes

mendaomn:
## Description

This PR makes it so the frontend redirects unauthenticated or unauthorized users to the authorization server. As it is now, it basically replicates the behavior that [was previously on the backend](https://github.com/aws-samples/pcluster-manager/pull/216)

## Changes

- Add `/get-app-config` route to share environment variables (and possibly other configs) to the client
- Add decorator over the `request` function to intercept 401/403 errors and redirect the user to the authorization server

## How Has This Been Tested?

- Manually, creation and deletion of a cluster
- Jest and Pytest

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -154,7 +154,7 @@ def get_scopes_list():
     return SCOPES_LIST + " openid"
   return SCOPES_LIST
 
-def get_redirect_url():
+def get_redirect_uri():
   return f"{SITE_URL}/login"
   
 # Local Endpoints
@@ -165,10 +165,10 @@ def get_version():
 
 def get_app_config():
   return {
-    "auth_path": AUTH_PATH,
+    "auth_url": AUTH_URL,
     "client_id": CLIENT_ID,
     "scopes": get_scopes_list(),
-    "redirect_url": get_redirect_url()
+    "redirect_uri": get_redirect_uri()
   }
 
 def ec2_action():
@@ -654,7 +654,7 @@ def login():
     url = TOKEN_URL
     code_resp = requests.post(
         url,
-        data={"grant_type": grant_type, "code": code, "client_id": CLIENT_ID, "redirect_uri": get_redirect_url()},
+        data={"grant_type": grant_type, "code": code, "client_id": CLIENT_ID, "redirect_uri": get_redirect_uri()},
         auth=auth,
         headers={"Content-Type": "application/x-www-form-urlencoded"},
     )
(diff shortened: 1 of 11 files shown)


frontend/src/auth/handleNotAuthorizedErrors.ts
```diff
@@ -0,0 +1,31 @@
...
+          redirectToAuthServer(authPath, clientId, scopes, redirectUrl)
+          return Promise.reject(error)
+      }
+      return Promise.reject(error)
+    }
+  )
+}
+
+function redirectToAuthServer(authPath: string, clientId: string, scopes: string, redirectUrl: string) {
+  const url = `${authPath}/login?response_type=code&client_id=${clientId}&scope=${encodeURIComponent(scopes)}&redirect_uri=${redirectUrl}`
```
BarcoMasile, line 29: issue(blocking): the authPath should already contain the whole path to redirect to, so "/login" should be removed since not every OIDC provider has that endpoint suffix
I added a comment about the get_config() API so please check that to adjust this

api/PclusterApiHandler.py
```diff
@@ -162,13 +147,29 @@ def _wrapper_authenticated(*args, **kwargs):
...
+  
 # Local Endpoints
 
 
 def get_version():
     return {"version": API_VERSION, "enable_mfa": ENABLE_MFA == "true"}
 
+def get_app_config():
+  return {
+    "auth_path": AUTH_PATH,
```
BarcoMasile, line 168: issue(blocking): please use "AUTH_URL" instead of "AUTH_PATH", since the URL to redirect the user to, in order to perform the login" is in that env var

api/PclusterApiHandler.py
```diff
@@ -162,13 +147,29 @@ def _wrapper_authenticated(*args, **kwargs):
...
+def get_scopes_list():
+  if not SCOPES_LIST:
+    return "openid"
+  elif "openid" not in SCOPES_LIST:
+    return SCOPES_LIST + " openid"
+  return SCOPES_LIST
 
+def get_redirect_url():
+  return f"{SITE_URL}/login"
+
```
BarcoMasile, line 159: nitpick(non-blocking):
thank you for providing this function.
I would use "redirect_uri" instead of redirect_url (also in the rest of the code),
to better comply with OAuth2 terminology (I don't think it's going to cause confusion)
but feel free to discard this comment since it's a very small nitpick

api/PclusterApiHandler.py
ermanno: Just to we are clear, `f'{AUTH_URL}'` is `f'{AUTH_PATH}/login'`, you are saying that the url should contain the `login` bit @BarcoMasile

BarcoMasile: I'm saying the opposite
"/login" should not be added manually by the frontend (see my other [comment](https://github.com/aws-samples/pcluster-manager/pull/220#discussion_r930046830)) since it is not mandatory that OIDC provider auth urls end in "/login"

ermanno: But then `$AUTH_PATH` should be correct here, no?

ermanno: Synced up offline. The point here was that `/login` is used only bi Cognito (our default), but not by other providers, and we should use the value provided by the customer in the ENV varibale as is.

ermanno (approved):

PR #225, opened 2022-07-27
Use "state" parameter for OAuth2 initiated login

BarcoMasile:
## Use `state` parameter for OAuth2 initiated login
This PR introduces the use of the state parameter on the frontend, populated with a random string.
This is part of the [RFC for OAuth2](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1)

State parameter is currently [supported by Cognito](https://docs.aws.amazon.com/cognito/latest/developerguide/login-endpoint.html).

### Reason
Right now we don't actually make use of the state parameter but introducing it allows us to better support other OIDC providers like Okta that request it as mandatory (even if the RFC puts it as recommended).

## Changes
- added `/frontend/tsconfig.tsbuildinfo` to .gitignore since it's used to check issues with typescript before doing the PR
- added a function to generate a simple random string to set the `state` parameter
- adjusted a test to reflect this change

## How Has This Been Tested?
This has been tested via Postman initially and then via a local deployment of PCM.
Tests have been run both for the frontend and the backend.

### DIFF
--- a/frontend/src/auth/__tests__/handleNotAuthorizedErrors.test.ts
+++ b/frontend/src/auth/__tests__/handleNotAuthorizedErrors.test.ts
@@ -40,10 +40,14 @@ describe('given the application config', () => {
         handleNotAuthorizedErrors(mockAppConfig)(mockRequestPromise).catch(() => {
             expect(global.window.location.replace).toHaveBeenCalledTimes(1)
             expect(global.window.location.replace).toHaveBeenCalledWith(
-                expect.stringMatching(/some-url\?response_type=code&client_id=some-id&scope=some-list&redirect_uri=some-uri&state=[a-zA-Z0-9]{1,32}/)
+                expect.stringContaining('some-url?response_type=code&client_id=some-id&scope=some-list&redirect_uri=some-uri')
             )
-          })
+            // 'as any' to avoid breaking npx tsc -p tsconfig.json
+            const urlString = `'${(global.window.location.replace as any).mock.calls[0][0]}`;
+            const queryParams = urlString.split('?')[1].split('&').map(keyvalue => keyvalue.split('='))
+            expect(queryParams.filter(([key, value]) => key === 'state').length).toBeGreaterThan(0)
 
+          })
         rejectPromise({ response: { status: 401 } })
       })
     })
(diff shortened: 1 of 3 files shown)


frontend/src/auth/__tests__/handleNotAuthorizedErrors.test.ts
```diff
@@ -39,9 +39,15 @@ describe('given the application config', () => {
       it('should redirect to the authorization server', () => {
         handleNotAuthorizedErrors(mockAppConfig)(mockRequestPromise).catch(() => {
             expect(global.window.location.replace).toHaveBeenCalledTimes(1)
-            expect(global.window.location.replace).toHaveBeenCalledWith('some-url?response_type=code&client_id=some-id&scope=some-list&redirect_uri=some-uri')
-          })
+            expect(global.window.location.replace).toHaveBeenCalledWith(
+                expect.stringContaining('some-url?response_type=code&client_id=some-id&scope=some-list&redirect_uri=some-uri')
+            )
+            // 'as any' to avoid breaking npx tsc -p tsconfig.json
```
mendaomn, line 45: There's no need to comment type assertions :)

frontend/src/auth/__tests__/handleNotAuthorizedErrors.test.ts
```diff
@@ -39,9 +39,15 @@ describe('given the application config', () => {
...
+            const urlString = `'${(global.window.location.replace as any).mock.calls[0][0]}`;
+            const queryParams = urlString.split('?')[1].split('&').map(keyvalue => keyvalue.split('='))
+            expect(queryParams.filter(([key, value]) => key === 'state').length).toBeGreaterThan(0)
```
mendaomn, line 48: suggestion: I think it would be a bit clearer and less error prone if we did this as follows:

```ts
const searchParams = new URL(urlString).searchParams
expect(searchParams.get('state')).toBeTruthy()
```

You'd simply need to set the mocked path to be a reasonable path, such as `http://somepath.com`

BarcoMasile: Unfortunately new URL(urlString) throws an exception with the "fake" url, I couldn't get it to work
so I had to do it manually

BarcoMasile: I couldn't get it to work even adding protocol and trying like "http://some.url"
new URL() throws an exception

frontend/src/auth/__tests__/handleNotAuthorizedErrors.test.ts
BarcoMasile: I'll remove it

mendaomn (approved):

PR #227, opened 2022-07-27
Move code for getting list of custom bootstrap actions to util.tsx

rkilpadi:
## Description

Details.tsx has some code that checks whether Slurm Accounting is active on a cluster. I moved this to utils so that it can be reused to check for other custom bootstrap actions (specifically Slurm REST API).

## How Has This Been Tested?

- Checked that the accounting tab was visible only when an accounting cluster was selected.
- Created tests for `getScripts()`

### DIFF
--- a/frontend/src/old-pages/Clusters/__tests__/util.test.tsx
+++ b/frontend/src/old-pages/Clusters/__tests__/util.test.tsx
@@ -1,0 +1,72 @@
+import { getScripts } from "../util";
+
+describe("Given a function to get script names of all custom actions", () => {
+    
+    describe("when a custom action to be run on node startup is provided", () => {
+        it("should return the name of the script", () => {
+            const scriptNames = getScripts({
+                OnNodeStart: {
+                    Script: 'https://www.website.com/test/directory/test-script.sh',
+                    Args: []
+                }
+            })
+            expect(scriptNames).toEqual(['test-script'])
+        })
+    })
+    
+    describe("when a custom action to be run on node configuration is provided", () => {
+        it("should return the name of the script", () => {
+            const scriptNames = getScripts({
+                OnNodeConfigured: {
+                    Script: 'https://www.website.com/test/directory/test-script.sh',
+                    Args: []
+                }
+            })
+            expect(scriptNames).toEqual(['test-script'])
+        })
+    })
+  
+    describe("when a custom action with arguments is provided", () => {
+        it("should return the names of all scripts", () => {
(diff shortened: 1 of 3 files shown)


frontend/src/old-pages/Clusters/util.tsx
```diff
@@ -27,3 +28,23 @@ export function selectCluster(clusterName: any, navigate: any)
   if(oldClusterName !== clusterName)
     setState(['app', 'clusters', 'selected'], name);
 }
+
+export function getScripts(customActions: [key: string]) {
```
mendaomn, line 32: Is it possibile that you meant the following?

```ts
customActions: {[key: string]}
```

frontend/src/old-pages/Clusters/util.tsx
```diff
@@ -27,3 +28,23 @@ export function selectCluster(clusterName: any, navigate: any)
...
+    if(getIn(customActions, [actionName, 'Script']))
+      allScripts.push(scriptName(getIn(customActions, [actionName, 'Script'])));
+    for(let arg of (getIn(customActions, [actionName, 'Args']) || []))
+    {
+      if(arg.length > 0 && arg[0] !== '-')
+        allScripts.push(scriptName(arg));
+    }
+  }
+  return allScripts
+}
```
mendaomn, line 50: This would a be a great opportunity to write a test for this function! Feel free to ping for help if needed!

frontend/src/old-pages/Clusters/__tests__/util.test.tsx
```diff
@@ -0,0 +1,25 @@
+import { getScripts } from "../util";
+
+describe("given custom actions", () => {
+    describe("when scripts are present", () => {
+        it("should return the script names", () => {
```
mendaomn, line 5: suggestion: this doesn't fully capture what the function does. (I know I sent you these lines, let's find something that can document the function)

Here's what I would do:

1. split into two different `it` tests, one per supported custom action. Meaning, one for the `OnNodeStart` and one for the `OnNodeConfigured` since those are the only ones we actually support
2. explicit what this means in terms of business logic, rather than implementation. For instance, you could write `given a custom action to be run before any node deployment, when a custom script is provided, it should return the name of the script`, same goes for the `OnNodeConfigured`
3. test corner cases: `when no custom script is provided` or `when the script should be run with args`

mendaomn:
Great job, LGTM

mendaomn (approved):

PR #228, opened 2022-07-28
refactor: move Actions buttons from split panel header to table header

tmscarla:
## Description

Currently in PCM we have a TypeScript error due to the fact that the header property of the `SplitPanel` component takes a string while we are passing a `<Header>` component instead.

This is not compliant with Polaris design system, thus we should move actions buttons from `SplitPanel` header to `Table` header.

## Changes
* Moved `Actions` component from `SplitPanel` to `Table` header

## How Has This Been Tested?
* Manually tested all action buttons with success
   * Cluster stopped / started
   * Shell
   * Filesystem
   * Edited / deleted


mtfranchetto (approved):

mtfranchetto:
@tmscarla Approved with a caveat: from what I see on the console all services with a table + actions always show the buttons in a disabled state when a resource is not selected. This is in contrast to what we do because we show those buttons only when a cluster is selected. Do we want to keep this behavior or switch to what console services do?

PR #231, opened 2022-08-02
`[Storage]` Remove MUI Slider, various chores.

charlesg3:
## Description

This PR removes the MUI slider so that we have one less dependency on a second UX framework.

## Changes

- Add a util clamp function
- Remove the MUI Slider, replace with an input that clamps value on blur
- Convert EFS throughput input to use same clamping function
- Add types to options and remove ts-expect-errors

## How Has This Been Tested?

Manually tested the Lustre Storage Capacity input with: letters, negative numbers, positive numbers, numbers out of step
Manually tested the EFS Throughput  input with: letters, negative numbers, positive numbers, numbers out of step

image

### DIFF
--- a/frontend/src/old-pages/Configure/Storage.tsx
+++ b/frontend/src/old-pages/Configure/Storage.tsx
@@ -14,7 +14,7 @@
 import * as React from 'react';
 import i18next from "i18next";
 import { Trans, useTranslation } from 'react-i18next';
-import { findFirst } from '../../util'
+import { findFirst, clamp } from '../../util'
 
 // UI Elements
 import {
@@ -24,12 +24,11 @@ import {
   FormField,
   Header,
   Input,
+  InputProps,
   Select,
   SpaceBetween,
   Toggle,
 } from "@awsui/components-react";
-
-import Slider from '@mui/material/Slider';
 
 // State
 import { getState, setState, useState, clearState } from '../../store'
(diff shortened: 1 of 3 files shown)


frontend/src/util.ts
```diff
@@ -70,4 +70,12 @@ function clusterDefaultUser(cluster: any) {
     "centos7": "centos"}[os]
 }
 
-export { getIn, swapIn, setIn, updateIn, findFirst, clusterDefaultUser }
+function clamp(num: number, min: number, max: number, step?: number): number
```
mendaomn, line 73: issue(non-blocking): Since this function is a `util` and will likely be used throughout the codebase we should write a test for this. This would also help in document the `step` argument which isn't typically included in the clamp implementations I'm used to.

Otherwise, you could use [lodash.clamp](https://lodash.com/docs/4.17.15#clamp) since I believe we already have `lodash` among the dependencies (it does lack the step option though)

frontend/src/old-pages/Configure/Storage.tsx
```diff
@@ -418,9 +422,7 @@ function EbsSettings({
             <FormField errorText = {volumeErrors}>
               <Input
                 disabled={editing}
-                style={{marginTop: 10}}
-                // @ts-expect-error TS(2322) FIXME: Type '"decimal"' is not assignable to type 'Type |... Remove this comment to see the full error message
-                type="decimal"
+                type={'decimal' as InputProps.Type}
```
mendaomn, line 425: issue(blocking): I don't see `decimal` among the supported types. They do list `decimal` as `inputMode`, did you mean to specify the input mode?

[reference](https://polaris.a2z.com/components/input/?tabId=api)

charlesg3: Good catch -- I believe I was more looking for `type='number'` but I updated that and added the input mode as well. :+1:

mendaomn (approved):

PR #214, opened 2022-07-23
`[StatusIndicator]` Convert the Status component to StatusIndicator.

charlesg3:
## Description

Convert the `Status`  component to use the Polaris `StatusIndicator` component (removing more MUI)

## How Has This Been Tested?

Manually by having several elements in each of various status states.

image

## References

https://polaris.a2z.com/components/status-indicator/?tabId=playground

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #232, opened 2022-08-02
`[Clusters/Scheduling]` Use updated status and remove more MUI.

charlesg3:
## Description

Use the new Status component ( https://github.com/aws-samples/pcluster-manager/pull/214 ) in the scheduling tab to remove move MUI.

## How Has This Been Tested?

By manually cycling jobs through the various states and validating their display:

image

image


frontend/src/components/Status.tsx
```diff
@@ -42,18 +42,25 @@ export default function Status({
...
-    "DELETE_FAILED": 'error',
-    "RUNNING": 'success',
-    "STOPPED": 'error',
-    "STOP_REQUESTED": 'in-progress',
-    "UPDATE_FAILED": 'error',
-    "UPDATE_IN_PROGRESS": 'in-progress',
-    "UNKNOWN": 'error',
-    "UPDATE_COMPLETE": 'success'};
+  const statusMap: Record<string, StatusIndicatorProps.Type> = {'CREATE_IN_PROGRESS': 'in-progress',
+    'CREATE_COMPLETE': 'success',
```
mtfranchetto, line 46: suggestion(non-blocking): it would be good to add a union type to the status so that who uses this component has the intellisense (and not just a `string` hint)

frontend/src/model.tsx
```diff
@@ -701,19 +689,19 @@ function SubmitJob(instanceId: any, user: any, job: any, callback: any, failure:
 }
 
 
-function JobInfo(clusterName: any, instanceId: any, user: any, jobId: any, callback: any, failure: any) {
+function JobInfo(_clusterName: any, instanceId: any, user: any, jobId: any, successCallback?: Callback, failureCallback?: Callback) {
```
mtfranchetto, line 692: question: why is `clusterName` here if it's not used?

frontend/src/model.tsx
```diff
@@ -340,14 +331,13 @@ function BuildImage(imageId: any, imageConfig: any, successCallback=null, errorC
...
       notify(`Error creating: ${imageId} - ${error.response.data.message}`, 'error');
-      // @ts-expect-error TS(2349) FIXME: This expression is not callable.
       errorCallback && errorCallback(error.response.data)
     }
     console.log(error.response.data)
   })
 }
 
-async function ListOfficialImages(region: any) {
+async function ListOfficialImages(region?: any) {
```
mtfranchetto, line 340: chore: we can type all `region` parameters as `string`

frontend/src/old-pages/Clusters/Scheduling.tsx
```diff
@@ -308,7 +281,7 @@ export default function ClusterScheduling() {
                 {
                     id: "state",
                     header: "state",
-                    cell: item => <Status status={(item as any).job_state}/>,
+                    cell: item => <Status status={(item as any).job_state as string}/>,
```
mtfranchetto, line 284: suggestion: it's better to avoid these explicit downcasts as they go against the framework, and it's not gonna help us realize if we mess up with types (since you're forcing a cast(. What you can do is to type `item` properly (or the jobs list directly) so that a cast is not needed.

frontend/src/old-pages/Configure/Create.tsx
```diff
@@ -35,7 +35,7 @@ function handleWarnings(resp: any) {
   if(!resp.validatonMessages)
     return;
 
-  resp.validatonMessages.forEach((message: any, i: any) => {
+  resp.validatonMessages.forEach((message: any) => {
```
mtfranchetto, line 38: typo: `validatonMessages` (also above) - is this code working? I don't see any other references to this variable in the codebase

frontend/src/components/Status.tsx
```diff
@@ -42,18 +42,25 @@ export default function Status({
...
+    'CONFIGURING': 'in-progress',
+    'COMPLETING': 'in-progress',
+    'COMPLETED': 'success',
+    'DELETE_IN_PROGRESS': 'in-progress',
+    'DELETE_FAILED': 'error',
+    'FAILED': 'error',
+    'PENDING': 'pending',
+    'RUNNING': 'success',
+    'STOPPED': 'error',
+    'SUCCESS': 'success',
```
mtfranchetto, line 58: issue: I see some problems with this approach
- we are extending the statuses every time we need to handle a new resource or a new state of an existing resource, and this can go on indefinitely (I think)
- we mix different statuses of different resources in the same dictionary, and it's very difficult to understand the entity to which a particular status belongs to

While we can solve these problems in many ways, one solution I propose is to extract the mapping logic from this component and put them in entity-specific mapping functions, so that we can test the logic easily

```typescript
function instanceToStatus(instance): StatusIndicatorProps.Type {
  // map instance status to component status type
}

function jobToStatus(job): StatusIndicatorProps.Type {
  // map job status to component status type
}
...
```

Let's have a quick chat if you have better ideas on how to solve this matter!

mtfranchetto (changes requested):

frontend/src/old-pages/Configure/Create.tsx
charlesg3: It looks like we weren't showing the user the creation warnings previously due to the check in the lines above.

With the fix to the typo I see the warnings from the create call displayed via the notification mechanism:

image

frontend/src/model.tsx
charlesg3: fixed.

frontend/src/components/Status.tsx
charlesg3: I ended up adding the statusMapOverrides option and leveraging it for instances which have a lower-case status.

charlesg3:
FYI -- I added a lot more typing to the `Instances.tsx` file in the last push when I converted to using a `statusMapOverride`

mtfranchetto (approved):

PR #197, opened 2022-07-19
Enable Slurm REST API

rkilpadi:
## Description

- Added Chef script to enable slurmrestd on a head node.
- Configured slurm.conf to enable JWT authentication.
- Automated JWT token generation.
- Configured nginx to allow HTTPS calls to the slurm REST API.
- Added Slurm REST API custom script to pcluster manager.

## How Has This Been Tested?

- Manually created clusters and ran curl commands from my local machine.

## References

- Added a tutorial on how to enable custom script and test cluster creation.
- https://slurm.schedmd.com/rest_api.html

## Work in Progress

- slurmdb calls to api do not work, but this is not needed for job submission.
- Still need to add support for different OSes (currently only works on alinux).

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #235, opened 2022-08-02
Add manual script entry as a new job submission option

rkilpadi:
## Description

Added an option to submit jobs by file upload or manual entry through a code editor. This is a proof of concept for my Slurm REST API script. Jobs entered through this method are submitted through the Slurm API rather than SSM.

Screen Shot 2022-08-16 at 6 35 17 PM

## Changes

- Added a new component JobField to simplify JobSubmitDialog.
- Added a new API endpoint to proxy requests to the Slurm REST API.
- Added new `script` submit option, which is only selectable when a cluster has the Slurm REST API Custom Bootstrap Action enabled. ~~(waiting on #227)~~
- Fixed a couple bugs/issues along the way
  - State did not clear when user inputted and then deleted a job property, which caused the job to fail
  - Error messages were never cleared
  - Refactored SubmitJob endpoint to return a Promise
  - Use translations in JobSubmitDialog.tsx

## How Has This Been Tested?

Manually submitted jobs with different properties and checked the request submitted to the Slurm API.

## References

#197
https://slurm.schedmd.com/rest_api.html#slurmV0038SubmitJob

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -271,6 +271,69 @@ def submit_job():
     return resp if type(resp) == tuple else {"success": "true"}
 
 
+def translate_job(request_body, user):
+    # format job parameters so that they are accepted by Slurm API
+    job_properties_dict = {
+        "job-name": "name",
+        "chdir": "current_working_directory",
+        "mem": "memory_per_node"
+    }
+
+    translated_job = {
+        "job": {
+           "environment": {
+                "PATH": "/bin:/usr/bin/:/usr/local/bin/:/opt/slurm/bin/"
+            },
+            "name": ""
+        }
+    }
+
+    request_body["chdir"] = f"/home/{user}"
+    for key in request_body:
+        if key == "command":
+            translated_job["script"] = request_body[key]
+        else:
+            translated_job["job"][job_properties_dict.get(key, key)] = request_body[key]
+
+    return translated_job
+
+
+def post_job_slurm_api(body, user, token, ip):
+    body_data = json.dumps(translate_job(body, user))
+
+    url = "https://"+ip+"/slurm/v0.0.36/job/submit"
+    headers = {
+        "Content-Type": "application/json",
+        "X-SLURM-USER-NAME": user,
+        "X-SLURM-USER-TOKEN": token
+    }
+
+    resp = requests.post(url=url, data=body_data, headers=headers, verify=False)
+    print("POST /slurm/v0.0.36/job/submit", resp.status_code, resp.reason, '\n')
+
+    error_dicts = resp.json().get('errors')
+    errors_text = '\n'.join([error_data.get("error") for error_data in error_dicts])
+
+    return {"errors": errors_text, "status_code": resp.status_code, "reason": resp.reason}
+
+
+def submit_job_script():
+    body = request.json
+    cluster_name = request.args.get("cluster_name")
+    instance_id = request.args.get("instance_id")
+    region = request.args.get("region")
+    user = request.args.get("user", "ec2-user")
+
+    ec2 = boto3.resource("ec2", region_name=region)
+    instance = ec2.Instance(instance_id)
+    ip = instance.public_dns_name
+
+    client = boto3.client("secretsmanager")
+    jwt_token = client.get_secret_value(SecretId="slurm_token_"+cluster_name)["SecretString"]
+
+    return post_job_slurm_api(body, user, jwt_token, ip)
+
+
 def _price_estimate(cluster_name, region, queue_name):
     config_text = get_cluster_config_text(cluster_name, region)
     config_data = yaml.safe_load(config_text)
--- a/frontend/src/model.tsx
+++ b/frontend/src/model.tsx
@@ -21,6 +21,7 @@ import identityFn from 'lodash/identity';
 import { getAppConfig } from './app-config';
 
 // Types
+import { Job } from './types/jobs';
 type Callback = (arg?: any) => void;
 
 const axiosInstance = axios.create({
@@ -680,25 +681,33 @@ function CancelJob(instanceId: any, user: any, jobId: any, callback?: Callback) 
   })
 }
 
-function SubmitJob(instanceId: any, user: any, job: any, successCallback?: Callback, failureCallback?: Callback) {
+async function SubmitJob(instanceId: string, user: string, job: Job): Promise<string> {
   const region = getState(['app', 'selectedRegion']) || getState(['aws', 'region']);
   let url = `manager/submit_job?instance_id=${instanceId}&user=${user || 'ec2-user'}&region=${region}`
-  request('post', url, job).then((response: any) => {
-    if(response.status === 200) {
-      console.log(response.data)
-      successCallback && successCallback(response.data)
-    }
-  }).catch((error: any) => {
-    if(error.response)
-    {
-      failureCallback && failureCallback(error.response.data.message)
-      console.log(error.response)
+  try {
+    const { data } = await request('post', url, job);
+    return data?.message || "";
+  } catch (error: any) {
+    if(error.response) {
       notify(`Error: ${error.response.data.message}`, 'error');
     }
-    console.log(error)
-  })
+    throw error;
+  }
 }
 
+async function SubmitJobScript(clusterName: string, instanceId: string, user: string, job: Job): Promise<string> {
+  const region = getState(['app', 'selectedRegion']) || getState(['aws', 'region']);
+  let url = `manager/submit_job_script?cluster_name=${clusterName}&instance_id=${instanceId}&user=${user || 'ec2-user'}&region=${region}`
+  try {
+    const { data } = await request('post', url, job);
+    return data?.errors || "";
+  } catch (error: any) {
+    if(error.response) {
+      notify(`Error: ${error.response.data.message}`, 'error');
+    }
+    throw error;
+  }
+}
 
 function JobInfo(instanceId: any, user: any, jobId: any, successCallback?: Callback, failureCallback?: Callback) {
   const region = getState(['app', 'selectedRegion']) || getState(['aws', 'region']);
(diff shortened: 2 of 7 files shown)


api/PclusterApiHandler.py
```diff
@@ -271,6 +271,50 @@ def submit_job():
     return resp if type(resp) == tuple else {"success": "true"}
 
 
+def translate_job(body, user):
```
tmscarla, line 274: suggestion: I found the whole function (naming in particular) a little bit convoluted :)

What do you think of a refactoring? Something like:

```
def translate_job(body, user):
    job_properties = {
        "job-name": "name",
        "chdir": "current_working_directory"
    }
    
    translated_job = {
        "job": {
            "environment": {
                "PATH":"/bin:/usr/bin/:/usr/local/bin/"
            }
        }
    }

    body['chdir'] = f"/home/{user}"    
    for key in body:
        if key == 'command':
            translated_job['script'] = body['command']
        else:
            translated_job['job'][job_properties.get(key, key)] = body[key]

    return translated_job
```

api/PclusterApiHandler.py
```diff
@@ -271,6 +271,50 @@ def submit_job():
...
+    # format job parameters so that they are accepted by Slurm API
+    job_properties = {
+        'job-name': 'name',
+        'chdir': 'current_working_directory'
+    }
+
+    body['chdir'] = '/home/'+user
```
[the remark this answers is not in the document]
rkilpadi: I think so, `get_dcv_session()` on line 456 uses the same path.

api/PclusterApiHandler.py
rkilpadi: You're right, it's isn't very clear -  did my best to clean it up

api/PclusterApiHandler.py
```diff
@@ -271,6 +271,50 @@ def submit_job():
...
+    return new_body
+
+
+def submit_job_script():
+    body = request.json
+    cluster_name = request.args.get("cluster_name")
+    instance_id = request.args.get("instance_id")
+    region = request.args.get("region")
+    user = request.args.get("user")
+    body_data = json.dumps(translate_job(body, user)) if body else None
```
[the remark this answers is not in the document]
rkilpadi: Empty commands not being handled properly was a pre-existing issue. In my latest commit I catch empty commands before they are submitted to the backend - let me know if you have feedback on how I did this.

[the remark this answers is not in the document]
rkilpadi: Empty commands not being handled properly was a pre-existing issue. In my latest commit I catch empty commands before they are submitted to the backend. Let me know if you have feedback on how I did this.

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
```diff
@@ -145,55 +146,137 @@ function JobCostEstimate() {
...
+  header,
+  description,
+  placeholder,
+  property,
+  disabled=false
+}: JobFieldProps) {
+
+  const setStateIfNotEmpty = (detail: any) => {
+    !detail || detail === '' ? clearState([...jobPath, property]) : setState([...jobPath, property], detail);
+  }
```
mendaomn, line 167: issue(types): you can properly type `detail` by using the types exported by the library: `InputProps.ChangeDetail`

issue(best-practice): we should always rely on `useCallback`, when applicable, for functions that we create inside the component and then pass down as props. See [here](https://reactjs.org/docs/hooks-reference.html#usecallback)

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
```diff
@@ -145,55 +146,137 @@ function JobCostEstimate() {
...
-  let jobName = useState([...jobPath, 'job-name']);
-  let chdir = useState([...jobPath, 'chdir']);
-  let nodes = useState([...jobPath, 'nodes']);
-  let ntasks = useState([...jobPath, 'ntasks']);
-  let mem = useState([...jobPath, 'mem']);
-  let command = useState([...jobPath, 'command']);
-  let wrap = useState([...jobPath, 'wrap']) || false;
+  let jobType: string = useState([...submitPath, 'job-entry']);
+
+  React.useEffect(() => jobTypeSelect('command'), []);
```
mendaomn, line 201: Do we need this state to be kept globally? If not you should keep the state only inside this component (via `React.useState`). This would also allow you to initialize state via the appropriate initializer (see [here](https://reactjs.org/docs/hooks-reference.html#usestate))

This should also avoid potential UI glitches that may affect low powered devices (while the state is not initialized the UI may be different, than change after you fire the `useEffect`)

api/PclusterApiHandler.py
```diff
@@ -271,6 +271,57 @@ def submit_job():
...
+    url = "https://"+ip+"/slurm/v0.0.36/job/submit"
+    headers = {
+        "Content-Type": "application/json",
+        "X-SLURM-USER-NAME": user,
+        "X-SLURM-USER-TOKEN": jwt_token
+    }
+
+    resp = requests.post(url=url, data=body_data, headers=headers, verify=False)
+    return resp if type(resp) == tuple else {"success": "true"}
+
```
mendaomn, line 323: issue(blocking): This handler does not have a test. In order to make it easier to test, you could extract the actual business logic into a separate function and test that:

```python
def actual_business_logic(body, user, token, ip):
 # generate body_data
 # request.post
 # generate response
```

You can then test the request is performed correctly (url, headers and body), and check the response is the one you expect (either the actual response or the fallback one)

frontend/src/model.tsx
```diff
@@ -698,6 +698,24 @@ function SubmitJob(instanceId: any, user: any, job: any, successCallback?: Callb
...
+  }).catch((error: any) => {
+    if(error.response)
+    {
+      failureCallback && failureCallback(error.response.data.message)
+      console.log(error.response)
+      notify(`Error: ${error.response.data.message}`, 'error');
+    }
+    console.log(error)
+  })
+}
```
mendaomn, line 718: suggestion: this should be refactored as it was done for `ListOfficialImages` and `ListClusters`. Main points to be addressed:

- SubmitJobScript should return a Promise that resolves with the actual data. This is needed so that we can then use `react-query` for data fetching
- instead of using callbacks, we can subscribe to the returned promise and simply call `.then` and `.catch` if needed

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
```diff
@@ -145,55 +146,137 @@ function JobCostEstimate() {
...
+
+  const jobTypeHeader = {
+    'command': "Command",
+    'file': "Script Path"
+  }[jobType]
+
+  const jobTypeDescription = {
+    'command': "The command to run as a part of this job.",
+    'file': "Path to the script to run."
+  }[jobType]
```
mendaomn, line 235: issue(blocking): Careful, by hardcoding strings this way we would not be able to support different languages in the app. Please use translation tokens instead, you can see [here an example](https://github.com/aws-samples/pcluster-manager/pull/175/commits/fdc6b77987c87a26f51dbc8da5d371d95ef80601)

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
```diff
@@ -145,55 +146,137 @@ function JobCostEstimate() {
...
+
+  const submitJobIfEntered = () => {
+    const entry = (getState([...jobPath, 'command']) || '').replace(/\n| /gm, '');
+    if(entry === '#!/bin/bash' || entry === '') {
+      setState([...submitPath, 'error'], "Error: Job is empty");
+    } else {
+      submitJob();
+    }
+  }
```
mendaomn, line 250: See useCallback comment 😉

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
```diff
@@ -205,105 +288,84 @@ export default function JobSubmitDialog({
...
-              placeholder="0"
-            />
-          </FormField>
-        </SpaceBetween>
+        <JobField
+          header="Number of tasks"
+          description="Number of tasks for job [optional]"
+          placeholder="0"
+          property='ntasks'
+        />
```
mendaomn, line 326: See translations comment 😉

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
```diff
@@ -205,105 +288,84 @@ export default function JobSubmitDialog({
           <SpaceBetween direction="horizontal" size="xs">
             <Button onClick={cancel}>Cancel</Button>
             {/* @ts-expect-error TS(2322) FIXME: Type '{ children: string; loading: any; onClick: (... Remove this comment to see the full error message */}
```
mendaomn, line 290: issue(types): Please remove the `autoFocus` property to address this error

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
```diff
@@ -205,105 +288,84 @@ export default function JobSubmitDialog({
...
         <QueueSelect />
 
-        <Toggle checked={wrap} onChange={({detail}) => enableWrap(!wrap)}>Run a Command (instead of script)</Toggle>
+        <RadioGroup 
+          onChange={({detail}) => {jobTypeSelect(detail.value);}}
+          value={jobType}
+          items={[
+            {value: "command", label: "Run a command"},
+            {value: "file", label: "Run a script on the head node"},
+            {value: "script", label: "Enter script manually", disabled: !isSlurmApiEnabled, description: scriptSelectDescription}
```
mendaomn, line 346: See translations comment 😉

mendaomn (changes requested):

mendaomn (approved):

frontend/src/model.tsx
```diff
@@ -680,25 +681,33 @@ function CancelJob(instanceId: any, user: any, jobId: any, callback?: Callback)
   })
 }
 
-function SubmitJob(instanceId: any, user: any, job: any, successCallback?: Callback, failureCallback?: Callback) {
+async function SubmitJob(instanceId: string, user: string, job: Job): Promise<string> {
```
mtfranchetto, line 684: praise: great job on refactoring this function!

frontend/src/old-pages/Clusters/JobSubmitDialog.tsx
```diff
@@ -10,9 +10,13 @@
 // limitations under the License.
```
mtfranchetto, line 10: suggestion(non-blocking): there is a bit of logic in this component not covered by tests. It can be added in another PR since this one already got a lot of iterations

mtfranchetto (approved):

PR #239, opened 2022-08-03
Managed Script for ODCR's

sean-smith:
## Description

This adds a script to make the process of using a ODCR with pcluster easier.

## Changes

Create new `odcr.sh` managed post install script.

## How Has This Been Tested?

Tested locally, verified cluster creates and instances are launched from the ODCR.

## References

https://docs.aws.amazon.com/parallelcluster/latest/ug/launch-instances-odcr-v3.html


docs/content/02-tutorials/07-odcr.md
```diff
@@ -0,0 +1,94 @@
++++
+title = "d. On-Demand Capacity Reservation 🤖"
```
charlesg3, line 2: I think this needs a different letter

docs/content/02-tutorials/07-odcr.md
```diff
@@ -0,0 +1,94 @@
...
+
+    ![On-Demand Capacity Reservation](odcr/post-install.png)
+
+    Fill in the parameters accordingly:
+
+    | **Parameter**                   | **Description**                                                                                                                                                         |
+    |---------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+    | Queue Name                      | Name of the queue you want to use the capacity reservation with.                                                                                                        |
+    | Compute Resource                | Name of the compute resource within the queue, this is typically `[queue_name]-[instance_type]` without the `.` in the instance type. i.e. `queue0-p4d24xlarge`         |
+    | Capacity Reservation Group Name | This is the name of the Capacity Group you created in [Step 1](https://docs.aws.amazon.com/parallelcluster/latest/ug/launch-instances-odcr-v3.html#odcr-create-cluster) |
```
charlesg3, line 22: It is also possible to do ODCR with a reservation ID so we should consider parameterizing that

charlesg3 (changes requested):
A few minor comments inlined

docs/content/02-tutorials/07-odcr.md
sean-smith: ah yes, I think that's actually better since it's simpler for the customer to just put in the reservation id

charlesg3: I was thinking that we would support both -- by specifying the type (and default it to `CapacityReservationResourceGroupArn`) so that it can be changed. Both are needed depending on the type of reservation.

charlesg3:
ODCR support is in progress on PC so we should leverage that in the next release

sean-smith:
Closing this in favor of next pcluster release

PR #240, opened 2022-08-03
Freeze requirements.txt

mendaomn:
## Description

This PR locks dependencies to the specific version we are currently using.

**Please note:**
The requirements.txt file is currently missing `flask` (and may be missing something else). Also we currently list `pyyaml` which I don't think is used anywhere. We should address this issues

## Changes

- change `requirements.txt` via `pip freeze`

## How Has This Been Tested?

- running `pip install -r requirements.txt` yields `Requirement already satisfied` for every dependency listed in the requirements file
- cleaned up the virtualenv using `pip uninstall -y -r <(pip freeze -r requirements.txt)` and reinstalled dependencies via `pip install -r requirements.txt`. Installed versions match the content of the frozen requirements.txt

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,8 +1,8 @@
-flask_restful
-flask_cors
-boto3
-requests
-python-jose
-pyyaml
-pytest
-pytest-mock
+Flask-RESTful==0.3.9
+Flask-Cors==3.0.10
+boto3==1.24.30
+requests==2.28.1
+python-jose==3.3.0
+PyYAML==6.0
+pytest==7.1.2
+pytest-mock==3.8.2


BarcoMasile:
question:
From my `pip freeze` I see many more dependencies that seem important, like
```
urllib3==1.26.9
Werkzeug==2.1.2
Jinja2==3.1.2
```
and many other.

Should we include those as well, since they could change in the future, or should we rely on the fact they're (probably) transitive dependencies on the one mentioned?

BarcoMasile (approved):

mendaomn:
> question: From my `pip freeze` I see many more dependencies that seem important, like
> 
> ```
> urllib3==1.26.9
> Werkzeug==2.1.2
> Jinja2==3.1.2
> ```
> 
> and many other.
> 
> Should we include those as well, since they could change in the future, or should we rely on the fact they're (probably) transitive dependencies on the one mentioned?

Let's lock the currently defined dependencies in the `requirements.txt`. The others are transitive and can be ignored for now

PR #241, opened 2022-08-03
fix: on missing idtoken or accesstoken return 401

BarcoMasile:
## Description
Fixed a small bug on missing id_token in requests
It caused a 500 in the get_identity_() controller function

## How Has This Been Tested?
This has been tested in a deployed environment.

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -561,7 +561,7 @@ def get_identity():
 
     access_token = request.cookies.get("accessToken")
     id_token = request.cookies.get("idToken")
-    if not access_token:
+    if not (access_token and id_token):
         return {"message": "No access token."}, 401
     try:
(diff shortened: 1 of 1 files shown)


ermanno (approved):

api/PclusterApiHandler.py
```diff
@@ -561,7 +561,7 @@ def get_identity():
 
     access_token = request.cookies.get("accessToken")
     id_token = request.cookies.get("idToken")
-    if not access_token:
+    if not (access_token and id_token):
```
mendaomn, line 564: Please follow this PR up with a test case for this. You can look at `test_pcluster_api_handler.py` and `test_authenticate.py` to see how cookies have been handled/mocked in tests

BarcoMasile: I will, thanks!

PR #244, opened 2022-08-04
`[types]` Add global types from ParallelCluster and weave through code.

charlesg3:
## Description

Add the global types from AWS ParallelCluster (and Slurm)

## Changes

Leverage the global types to take advantage of type checking.

## How Has This Been Tested?

Manual testing.

## References

https://github.com/aws/aws-parallelcluster/tree/develop/api/spec/smithy/model


frontend/src/old-pages/Configure/Cluster.tsx
```diff
@@ -258,7 +258,7 @@ function VpcSelect() {
           <Select
             disabled={editing}
             // @ts-expect-error TS(2322) FIXME: Type '{ label: JSX.Element; value: any; }' is not ... Remove this comment to see the full error message
-            selectedOption={vpcToDisplayOption(findFirst(vpcs, (x: any) => {return x.VpcId === vpc}))}
+            selectedOption={vpcToDisplayOption(findFirst(vpcs, x => x.VpcId === vpc))}
```
mendaomn, line 261: suggestion: can we apply the same fix you did `JobSumitDialog.tsx` to the `itemToOption` function?

frontend/src/old-pages/Configure/Components.tsx
```diff
@@ -231,7 +238,7 @@ function CustomAMISettings({
       })
     }
 
-  const toggleCustomAmi = (event: any) => {
+  const toggleCustomAmi = (_e: any) => {
```
mendaomn, line 241: suggestion: I think here you can remove `_e` entirely, or otherwise we should type it appropriately with `NonCancelableEventHandler<ToggleProps.ChangeDetail>` which should be exported by `@awsui`

frontend/src/old-pages/Configure/Storage.tsx
```diff
@@ -704,7 +704,7 @@ function Storage() {
               <Select
                 disabled={editing}
                 placeholder="Please Select a Filesystem Type"
-                selectedOption={storageType && itemToDisplayIconOption(findFirst(storageTypes, (s: any) => {return s[0] === storageType}))}
+                selectedOption={storageType && itemToDisplayIconOption(findFirst(storageTypes, s => s[0] === storageType) || ['', '', ''])}
```
mendaomn, line 707: issue(non-blocking): by using `['', '' ,'']` as default, wouldn't we pass a possibly non-existing item to `selectedOption`? Would that cause any glitch?

frontend/src/types/constants.ts
```diff
@@ -0,0 +1,177 @@
...
+  Starting = "STARTING",  // works only with Slurm
+  Running = "RUNNING",  // works only with Slurm
+  Protected = "PROTECTED",  // works only with Slurm
+  StopRequested = "STOP_REQUESTED",  // works only with Slurm
+  Stopping = "STOPPING",  // works only with Slurm
+  Stopped = "STOPPED",  // works only with Slurm
+  Unknown = "UNKNOWN",  // works only with Slurm
+  Enabled = "ENABLED",  // works only with AWS Batch
+  Disabled = "DISABLED",  // works only with AWS Batch
+}
```
mendaomn, line 37: suggestion: I'm not sure I'm getting what's happening here. Are you merging two different types? If so, I'd find a union type (or other type merging strategy) a bit more robut.

I'm thinking something along these lines:

```ts
export type ComputeFleetStatus = SlurmFleetStatus | AWSBatchFleetStatus
```

This is [not easily done](https://stackoverflow.com/a/55827534) with `enum`s, but there could be a solution. 

Probably union types are the better choice here, since we're are basically defining types, but I don't have strong opinions against enums (however here's a [fairly short read against enums](https://fettblog.eu/tidy-typescript-avoid-enums/) 😁 )

frontend/src/types/global.d.ts
```diff
@@ -0,0 +1,387 @@
...
+    Ec2AmiState,
+    ClusterStatus,
+    ValidationLevel,
+    CloudFormationStackStatus,
+    CloudFormationResourceStatus,
+    JobStateCode,
+} from './enums'
+
+
+export {}
```
mendaomn, line 27: question: Is this somehow needed or just a leftover?

frontend/src/types/global.d.ts
```diff
@@ -0,0 +1,387 @@
...
+  type ValidationMessages = ConfigValidationMessage[]
+
+  type ValidationLevel = ValidationLevel
+  type CloudFormationStackStatus = CloudFormationStackStatus
+  type StackStatus = CloudFormationStackStatus
+
+  type CloudFormationResourceStatus = CloudFormationResourceStatus
+  type StackEventStatus = CloudFormationResourceStatus
+
+  // Instances
```
mendaomn, line 228: Spoiler for what's coming? :P

frontend/src/types/global.d.ts
```diff
@@ -0,0 +1,387 @@
...
+    ValidationLevel,
+    CloudFormationStackStatus,
+    CloudFormationResourceStatus,
+    JobStateCode,
+} from './enums'
+
+
+export {}
+
+declare global {
```
mendaomn, line 29: issue(non-blocking): Global types are ok, yet I would rather split this big file into smaller files according to their domain (e.g Instance, Image, Jobs, Logs, and so on)

The reasoning behind it would be to have a clearer separation between what part of the codebase has ownership over what types.

To give a maybe oversimplified example: if we got rid of the whole `Images` section, we could forget to remove the related types.

Thoughts on this?

frontend/src/old-pages/CustomImages/CustomImages.tsx
```diff
@@ -43,22 +43,18 @@ const imageBuildPath = ['app', 'customImages', 'imageBuild'];
 const selectCustomImagesList = (state: any) => state.customImages.list;
 
 function CustomImagesList() {
-  const images = useSelector(selectCustomImagesList) || [];
-  const open = useState([...imageBuildPath, 'dialog']);
+  const images: ImageInfoSummary[] = useSelector(selectCustomImagesList) || [];
```
mendaomn, line 46: FYI: I slight better way to do this would be to correctly type your selector. (The ideal way would be to type the Store itself, but we'll get there eventually)

The reason why it is slightly better to let TS infer the type for your `const images` is that if you were to change or inadvertently break the selector, it would yell at you in the rest of the code :) 

```ts
const selectCustomImagesList = (state: any) : ImageInfoSummary[] => state.customImages.list;
//...
const images = useSelector(selectCustomImagesList) || [];
```

frontend/src/old-pages/Configure/Cluster.tsx
charlesg3: Will fix in separate PR

frontend/src/old-pages/Configure/Storage.tsx
charlesg3: Ah I think this is probably better resolved by using `!` (which I just learned about) stating that we will have a match (which fixes the type error)

frontend/src/types/constants.ts
charlesg3: This is actually how it's defined directly from the types in ParallelCluster. So I might prefer to keep things consistent with that -- I'm happy to break it out which might make our typing easier in certain cases (?), could also be broken out when needed.

frontend/src/types/global.d.ts
charlesg3: Based on what I learned about type declarations, I believe this is needed to make the file a module. I added a comment.

frontend/src/types/global.d.ts
charlesg3: Some of the types are common, but that can also go in its own separate file.

I wasn't able to figure out a good way to co-locate the constants, so in some ways those would also need to reflect the same structure -- which might make it more inconvenient from the consumption side.

I'm not sure if we can / should get rid of something like the Images section -- since this is a copy of the types in PC I would expect it to only expand for the life of PC3.x -- though perhaps they should be put into some PC specific and then Slurm specific?

I'm open to whatever, though might do that re-org as a separate pass (feel free to assign the work to me)

frontend/src/types/global.d.ts
charlesg3: I just double-checked and this might not be needed so will try removing it.

frontend/src/types/global.d.ts
charlesg3: :P -- removed

frontend/src/types/global.d.ts
mtfranchetto: Regarding the global types I would keep things consistent: there are some places where you import constants directly (`import { ClusterStatus } from '../../types/constants'`) and some places where types get included automatically. I would pick the same method (I prefer importing modules since it's more explicit and easier to track down but no strong preference here)

Regarding the splitting: I would keep an auto-generated-like file **only** if we can run a command (for example if PC releases a new version) and get the updated types with no manual intervention, so we can "forget" about types. Otherwise, it will be better to do even a rough separation between those types, like:
- `base` types
- `jobs`
- `cluster`
- `instances`
- `images`
- and so on

If you choose the first approach (keep everything in the same file) I encourage you to write that part and the related commands inside the readme/package.json, so a developer could use it in the future.

frontend/src/old-pages/Clusters/Instances.tsx
```diff
@@ -173,49 +128,49 @@ export default function ClusterInstances() {
         {
             id: "id",
             header: "Id",
-            cell: (instance: Instance) => <Link external externalIconAriaLabel="Opens in a new tab" href={`${consoleDomain(region)}/ec2/v2/home?region=${region}#InstanceDetails:instanceId=${instance.instanceId}`}>{instance.instanceId}</Link>,
+            cell: instance => <Link external externalIconAriaLabel="Opens in a new tab" href={`${consoleDomain(region)}/ec2/v2/home?region=${region}#InstanceDetails:instanceId=${instance.instanceId}`}>{instance.instanceId}</Link>,
```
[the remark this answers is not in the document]
charlesg3: There seem to be an odd type issue and so we will push this chore to a systematic change across the project.

mtfranchetto (approved):

PR #252, opened 2022-08-05
Handle id token and access token attributes in get_identity

tmscarla:
## Description

Currently PCM retrieves `user_groups` and `username` claims from the access token, and store all the claims of the id token in the `attributes` field of the object returned by the `get_identity` function.

In order to avoid forcing the customer to specify which attributes will end up in the access token and which ones in the ID token, we have to refactor our current logic.

The idea is to get user attributes from the access token first and then from the id token, resolving all conflicts in favor of the id token.

## Changes

* `user_groups`, `username` and `email` claims are now retrieved from both access token and id token, giving higher priority to the latter (i.e. if a claim is available in both tokens, the id token claim will be returned in the identity object)
* dropped unused parameters in the returned identity object

## Changelog entry

* Handle both id token and access token claims in get_identity endpoint, giving higher priority to id token ones

## How Has This Been Tested?

* Run PCM locally pointing to Cognito user pool in my personal account setting the following environment variables:
```
export SECRET_ID
export AUTH_PATH
export USER_POOL_ID
export AUDIENCE
export CLIENT_ID
export CLIENT_SECRET
```
* Manually inspected `get_identity` output and verified that id token claims overrode access token claims

## PR Quality Checkilst

- [n/a] I removed hardcoded strings and used our `i18n` solution instead (see [here](https://github.com/aws-samples/pcluster-manager/pull/175/commits/fdc6b77987c87a26f51dbc8da5d371d95ef80601))

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -551,7 +551,9 @@ def get_instance_types():
     return {"instance_types": sorted(instance_types, key=lambda x: x["InstanceType"])}
 
 
-def _populate_identity(identity, decoded, claims):
+def _get_identity_from_token(decoded, claims):
+    identity = {"attributes": {}}
+
     if USER_ROLES_CLAIM in decoded:
         identity["user_roles"] = decoded[USER_ROLES_CLAIM]
     if "username" in decoded:
@@ -561,6 +563,7 @@ def _populate_identity(identity, decoded, claims):
       if claim in decoded:
         identity["attributes"][claim] = decoded[claim]
     
+    return identity
 
 def get_identity():
     if disable_auth():
@@ -571,13 +574,14 @@ def get_identity():
     if not (access_token and id_token):
         return {"message": "No access or id token."}, 401
     try:
-        identity = {"attributes": {}}
         claims = ["email"]
 
         decoded_access = jwt_decode(access_token)
-        _populate_identity(identity=identity, decoded=decoded_access, claims=claims)
+        identity_from_access_token = _get_identity_from_token(decoded=decoded_access, claims=claims)
         decoded_id = jwt_decode(id_token, audience=AUDIENCE, access_token=access_token)
-        _populate_identity(identity=identity, decoded=decoded_id, claims=claims)
+        identity_from_id_token = _get_identity_from_token(decoded=decoded_id, claims=claims)
+
+        identity = {**identity_from_access_token, **identity_from_id_token}
 
         if "username" not in identity:
             return {"message": "No username present in access or id token."}, 400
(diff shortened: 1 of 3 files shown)


api/PclusterApiHandler.py
```diff
@@ -551,9 +551,16 @@ def get_instance_types():
...
+def _populate_identity(identity, decoded, claims):
+    if USER_ROLES_CLAIM in decoded:
+        identity["user_roles"] = decoded[USER_ROLES_CLAIM]
+    if "username" in decoded:
+        identity["username"] = decoded["username"]
 
+    for key, value in decoded.items():
+        if key in claims:
+            identity["attributes"][key] = value
+
```
BarcoMasile, line 563: suggestion: since you have a list of claims to retrieve from the access/id token, here you could avoid cycling on all the elements in the JWT payload, and instead cycle over the claims list itself
```python3
for claim in claims:
  if claim in decoded:
    identity["attributes"][claim] = decoded[claim]
```

api/PclusterApiHandler.py
```diff
@@ -551,9 +551,16 @@ def get_instance_types():
     return {"instance_types": sorted(instance_types, key=lambda x: x["InstanceType"])}
 
 
-def _get_user_roles(decoded):
-    return decoded[USER_ROLES_CLAIM] if USER_ROLES_CLAIM in decoded else ["user"]
+def _populate_identity(identity, decoded, claims):
```
mtfranchetto, line 554: suggestion(non-blocking): for future contributions, I'd go with pure functions and immutability rather than modifying existing objects

api/tests/test_get_identity.py
```diff
@@ -6,10 +6,107 @@
 MOCK_IDENTITY_OBJECT = {"user_roles": ["user", "admin"], "username": "username", "attributes": {"email": "user@domain.com"}}
 
 @mock.patch.dict(os.environ,{"ENABLE_AUTH": "false"})
-def test_get_identity():
+def test_get_identity_auth_disabled():
```
mtfranchetto, line 9: chore: I see that we still import `unittest`, can we move to `pytest` already in these tests?

mtfranchetto (approved):

api/PclusterApiHandler.py
```diff
@@ -571,13 +574,14 @@ def get_identity():
...
         claims = ["email"]
 
         decoded_access = jwt_decode(access_token)
-        _populate_identity(identity=identity, decoded=decoded_access, claims=claims)
+        identity_from_access_token = _get_identity_from_token(decoded=decoded_access, claims=claims)
         decoded_id = jwt_decode(id_token, audience=AUDIENCE, access_token=access_token)
-        _populate_identity(identity=identity, decoded=decoded_id, claims=claims)
+        identity_from_id_token = _get_identity_from_token(decoded=decoded_id, claims=claims)
+
+        identity = {**identity_from_access_token, **identity_from_id_token}
```
tmscarla, line 584: praise: nice one :)

PR #256, opened 2022-08-17
Setup AWS credentials with Github Actions

mtfranchetto:
## Description

To access AWS resources inside a Github workflow you need to create new IAM roles under AWS and use a specific action that will retrieve temporary credentials to access your account.
This PR configures an AWS account to talk with the Github OIDC provider.

## Changes

- add a CFN template to automate roles and providers creation on AWS
- add a sample workflow to authenticate on AWS using the role created

## How Has This Been Tested?

- verified with [this run](https://github.com/aws-samples/pcluster-manager/actions/runs/2876412169/attempts/1) that a branch different from `main` cannot authenticate on AWS
- verified with [this run](https://github.com/aws-samples/pcluster-manager/actions/runs/2876412169) that the authentication works correctly


mendaomn (dismissed):

infrastructure/github-env-setup.yml
```diff
@@ -0,0 +1,66 @@
...
+        - PolicyName: UpdateLambdaFromPrivateECR
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Effect: Allow
+                Action:
+                  - lambda:UpdateFunctionCode
+                  - ecr:DescribeRepositories
+                Resource:
+                  - !Sub arn:aws:lambda:${AWS::Region}:${AWS::AccountId}:function:PClusterManagerFunction
```
ermanno, line 45: For these and other ARNs we should use the [pseudo parameter reference](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-partition) for the _partition_ as it might [change](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html) depending on the region.

ermanno (changes requested):

ermanno (approved):

maurizp (approved):

mendaomn (approved):

PR #272, opened 2022-08-22
Script to update a CFN stack

mtfranchetto:
## Description
Provides an easy way to update the infrastructure of a specific environment on which PCM is running.
The script is run locally because of the complex IAM policy required by the workflow (it may be added later).

See the infrastructure readme or the script itself for more info.

## Changes

- add a new parameter to specify the infrastructure bucket on CFN
  - if left empy it will use the public yaml files
  - if a user provides a custom bucket it will use the files inside it
- add a script to update a CFN stack

## How Has This Been Tested?

Launched `./infrastructure/update-environment-infra.sh demo` on the demo environment and verified the stack is updated correctly


mtfranchetto:
> I think we need to add the license header to the new files.
> 
> There's probably an ADR that we can add wrt the fact that we are opting for running the script locally (not sure it's worth it though)
> 
> LGTM otherwise

I don't see the license on the other YAML/sh files. Is it really needed?

PR #282, opened 2022-08-30
Add git-secrets automation

eantonin:
## Description

This PR introduces an automation for the setup of git-secrets forcing the installation of the relevant git hooks during the npm "prepare" phase, this works only if git-secrets is installed, so I added installation instructions in the CONTRIBUTING.md file

We may need to add more patterns in the future, but this can be done iteratively

## Changes

*    security - Force the installation of git hooks related to git-secrets at npm "prepare" phase
*    documentation (contributing) Adding instructions to install git-secrets
*    refactor - Adding shebang to bash script
*    refactor (ci) - Execute the git-secrets installation only if we are not on CI

## Changelog entry

N/A

## How Has This Been Tested?

- Run `npm install` without git-secrets installed -> fails
- Run `npm install` with git-secrets installed -> succeeds
  - Hooks installed
  - Tried to commit a secret -> committing fails with `[ERROR] Matched one or more prohibited patterns`
  - Commits without secrets -> succeeds

## References

[git-secrets-repo
](https://github.com/awslabs/git-secrets)

[npm scripts doc](https://docs.npmjs.com/cli/v8/using-npm/scripts)
## PR Quality Checklist
- [X] I checked that `npm run build` builds without any error

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -38,7 +38,8 @@
     "dev": "next dev",
     "export": "next build && next export -o build",
     "ts-reignore": "ts-migrate reignore . --sources='src/**/*'",
-    "ts-validate": "tsc -p tsconfig.json"
+    "ts-validate": "tsc -p tsconfig.json",
+    "prepare": "bash ./git-secrets-conf.sh"
   },
   "browserslist": {
     "production": [
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,9 +23,10 @@ reported the issue. Please try to include as much information as you can. Detail
 ## Contributing via Pull Requests
 Contributions via pull requests are much appreciated. Before sending us a pull request, please ensure that:
 
-1. You are working against the latest source on the *main* branch.
-2. You check existing open, and recently merged, pull requests to make sure someone else hasn't addressed the problem already.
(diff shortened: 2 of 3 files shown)


frontend/package.json
```diff
@@ -38,7 +38,8 @@
     "dev": "next dev",
     "export": "next build && next export -o build",
     "ts-reignore": "ts-migrate reignore . --sources='src/**/*'",
-    "ts-validate": "tsc -p tsconfig.json"
+    "ts-validate": "tsc -p tsconfig.json",
+    "prepare": "bash ./git-secrets-conf.sh"
```
BarcoMasile, line 42: suggestion(non-blocking): I would suggest a different name for this npm script, something that tells more about what it is, since "prepare" sounds to me a little to general.
Even something like "setup-git-secrets" would be fine, or something like that so that no doubts can arise.

BarcoMasile (approved):

frontend/package.json
eantonin: The "prepare" is a part of npm lifecycle see https://docs.npmjs.com/cli/v8/using-npm/scripts. This is to ensure the script it's executed at every npm install. So I cannot change that name.

BarcoMasile: Perfect then! Thanks!

PR #283, opened 2022-08-30
Fix authentication of admin not belonging to the user group

mtfranchetto:
## Description

Gives the ability for an admin which is not in the `users` group (Cognito) to use the application.
Previously, an administrator not belonging to the `users` group could not access PCM.
This is a requirement to integrate with a third-party provider, as we must not assume how users are created on that platform.

## Changes
- authenticate agains a set of groups rather than a single one
- include both the `admin` and `user` group for endpoints accessible by both

## Changelog entry

Fix permissions check against `admin` not belonging to the Cognito `user` group

## How Has This Been Tested?

Tested locally with Cognito:
- removed the `user` group from an admin account and verified it broke the application
- applied the changes to the backend
- verified the user can to browse PCM
- 
## PR Quality Checklist

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -137,11 +137,11 @@ def authenticate(groups):
     if ("guest" not in groups) and len(groups_granted) == 0:
         return abort(403)
 
-def authenticated(group="user"):
+def authenticated(groups={"user"}):
     def _authenticated(func):
         @functools.wraps(func)
         def _wrapper_authenticated(*args, **kwargs):
-            authenticate(group)
+            authenticate(groups)
             return func(*args, **kwargs)
 
         return _wrapper_authenticated
@@ -711,7 +711,7 @@ def _get_params(_request):
 
 
 class PclusterApiHandler(Resource):
-    method_decorators = [authenticated("user")]
+    method_decorators = [authenticated({"user", "admin"})]
 
     def get(self):
         # if re.match(r".*images.*logstreams/+", args["path"]):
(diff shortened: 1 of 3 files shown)


BarcoMasile (dismissed):

api/PclusterApiHandler.py
```diff
@@ -132,15 +132,16 @@ def authenticate(group):
...
 
-    if (group != "guest") and (group not in set(decoded.get(USER_ROLES_CLAIM, []))):
+    jwt_roles = set(decoded.get(USER_ROLES_CLAIM, []))
+    groups_granted = groups.intersection(jwt_roles)
+    if ("guest" not in groups) and len(groups_granted) == 0:
         return abort(403)
 
-
-def authenticated(group="user"):
+def authenticated(groups={"user"}):
```
eantonin, line 140: question(non-blocking): Why did you choose only user as default? using admin and user would make more sense in my opinion, and possibly would remove some code in the app.py

eantonin (approved):

api/PclusterApiHandler.py
mtfranchetto: I just kept the old behavior, but there are no places right now where the "default" is used (all invocations to `authenticated` are with parameters). Since I dislike these implicit behaviors, what I propose is to get rid entirely of the default parameter, and specify always a set of groups. What do you think? @eantonin

eantonin: Yeah, I think your proposal is the better solution in this case

mtfranchetto: Fixed!

eantonin (approved):

PR #287, opened 2022-09-02
Add linting and formatting

tmscarla:
## Description

Currently no tooling is available in the project to ensure that all changes follows the same formatting rules and avoids the use of some bad patterns. This PR aims to include in the project standard tools like ESLint (already bundled by Next), Prettier and lint-staged. 

Furthermore, it introduces [husky](https://github.com/typicode/husky) to the project for handling git hooks in a simplified and more organized manner.

Prettier rules have been chosen following JS/TS best practices.

## Changes

* ci - added linting (`eslint`) and formatting (`prettier`) to the project
* ci - leverage `lint-staged` to enhance performance applying linting to staged files only
* refactor - use husky for managing git hooks, including `git-secrets`

## Changelog entry

N/A

## How Has This Been Tested?

* Run `npx eslint --fix` on a subset of files, verified that files are linted correctly
* Run `npx prettier --write` on a subset of files, verified that files are formatted correctly
* Created a dummy .tsx file, added to staging and committed. Verified that linting and formatting is performed through git pre-commit hook set with `husky`
* Created a dummy .sh file with AWS credentials. Verified that on commit git-secret hooks are still applied.

## References

* https://nextjs.org/docs/basic-features/eslint#usage-with-other-tools
* https://paulintrognon.fr/blog/typescript-prettier-eslint-next-js

## PR Quality Checklist

### DIFF
--- a/frontend/.husky/commit-msg
+++ b/frontend/.husky/commit-msg
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+"$(dirname -- "$0")"/../scripts/git-secrets-command.sh --commit_msg_hook -- "$@"
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -1,3 +1,6 @@
 {
-  "extends": "next/core-web-vitals"
+  "extends": [
+    "next/core-web-vitals",
+    "prettier"
+  ]
 }
--- a/frontend/.husky/pre-commit
+++ b/frontend/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+npx lint-staged
--- a/frontend/.husky/pre-commit
+++ b/frontend/.husky/pre-commit
@@ -1,4 +1,5 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
+"$(dirname -- "$0")"/../scripts/git-secrets-command.sh --pre_commit_hook -- "$@"
 npx lint-staged
--- a/frontend/.husky/prepare-commit-msg
+++ b/frontend/.husky/prepare-commit-msg
@@ -1,0 +1,4 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+"$(dirname -- "$0")"/../scripts/git-secrets-command.sh --prepare_commit_msg_hook -- "$@"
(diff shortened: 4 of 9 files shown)


eantonin (commented):
question(blocking): how do you install husky? afaik it's not auto-installing see https://blog.typicode.com/husky-git-hooks-autoinstall/

frontend/.husky/commit-msg
```diff
@@ -0,0 +1,12 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+if [ -z "${CI}" ] || [ "${CI}" != true ]; then
+  if ! command -v git-secrets >/dev/null 2>&1; then
+      echo "git-secrets is not installed. Please visit https://github.com/awslabs/git-secrets#installing-git-secrets"
+      exit 1
+  fi
+
+  git-secrets --register-aws > /dev/null
```
eantonin, line 10: issue(non-blocking): there are two issues here:
1. I'm not really sure `git secrets --install` can be dropped, are you sure it's only creating git hooks files?
1. `git-secrets --register-aws > /dev/null` is not a part of the hook, is the configuration of git-secrets, so in my opinion should not be re-done for every hook, but should be done just at git secrets installation. So if you install husky at `prepare` I suggest to put `git-secrets --register-aws` (and maybe all the if) there

tmscarla:
> question(blocking): how do you install husky? afaik it's not auto-installing see https://blog.typicode.com/husky-git-hooks-autoinstall/

Good catch! I removed by mistake the entire following line in `package.json`:

```
"prepare": "bash ./git-secrets-conf.sh && cd .. && husky install frontend/.husky"
```

frontend/.husky/commit-msg
tmscarla: 1. Apparently yes: https://github.com/awslabs/git-secrets#options-for-install
2. Had the same doubt. This approach is proposed by this [article](https://www.codewithyou.com/blog/using-husky-and-gitsecrets-to-prevent-committing-passwords-and-other-sensitive-information-to-a-git-repository). I can try to put in prepare and test again.

BarcoMasile: suggestion(non-blocking): I'd suggest to group the shared logic under a single file to source in the various scripts

eantonin: I do not know if creates problems, but mind this line https://github.com/awslabs/git-secrets#options-for-install:~:text=Git%20only%20allows%20a%20single%20script%20to%20be%20executed%20per%20hook

eantonin (approved):

PR #290, opened 2022-09-02
Apply linting and formatting to the frontend codebase

tmscarla:
## Description

With [this PR](https://github.com/aws-samples/pcluster-manager/pull/287) we introduced linting and formatting through `ESlint` and `Prettier`. This PR aims at applying linting and formatting rules to the whole frontend codebase, as a starting point for having a uniform style across all JS/TS files.

## Changes

* Applied linting and formatting to the frontend codebase following rules specified in `.prettierc.json` and `.eslintrc.json`.

## Changelog entry

* N/A

## How Has This Been Tested?

* PCM built and run locally
* Run `npx eslint --fix`
* Run `npx prettier --write`
* Followed PR quality checklist to ensure that from a functional point of view no changes were made to the code

## References

* https://github.com/aws-samples/pcluster-manager/pull/287

## PR Quality Checklist

### DIFF
--- a/frontend/src/auth/constants.ts
+++ b/frontend/src/auth/constants.ts
@@ -1,1 +1,1 @@
-export const USER_ROLES_CLAIM = "user_roles"
+export const USER_ROLES_CLAIM = 'user_roles'
--- a/frontend/src/components/Status.tsx
+++ b/frontend/src/components/Status.tsx
@@ -8,163 +8,214 @@
 // or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 // OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 // limitations under the License.
-import { CloudFormationResourceStatus } from '../types/base'
-import { ClusterStatus, ClusterDescription, ClusterInfoSummary, ComputeFleetStatus } from '../types/clusters'
-import { InstanceState, Instance, EC2Instance } from '../types/instances'
-import { StackEvent } from '../types/stackevents'
-import { JobStateCode } from '../types/jobs'
-import React, { useCallback } from 'react';
-import { Trans } from 'react-i18next';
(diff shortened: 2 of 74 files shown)


frontend/src/auth/constants.ts
```diff
@@ -1 +1 @@
-export const USER_ROLES_CLAIM = "user_roles"
+export const USER_ROLES_CLAIM = 'user_roles';
```
eantonin, line 1: (question, non-blocking): Do we really want to enforce single quotes vs double? Do we really want to enforce colons?

frontend/src/components/Status.tsx
```diff
@@ -54,117 +78,144 @@ function ClusterStatusIndicator({cluster}: {cluster: ClusterInfoSummary | Cluste
...
   const statusMap: Record<InstanceState, StatusIndicatorProps.Type> = {
-    "pending": 'pending',
-    "running": 'success',
-    "shutting-down": 'loading',
-    "stopped": 'stopped',
-    "stopping": 'pending',
-    "terminated": 'stopped',
+    pending: 'pending',
+    running: 'success',
+    'shutting-down': 'loading',
```
eantonin, line 166: question(non-blocking): Why this needs quotes?

mendaomn: because it wouldn't be valid syntax otherwise 

```ts
const obj = {
  a-b: "anything" // syntax error
}
```

frontend/src/auth/constants.ts
tmscarla: They are both defaults Prettier options. Personally I would keep the single quotes and drop the semicolon :)

frontend/src/components/Status.tsx
eantonin: ok, at that point I'd put all the variables inside some single or double quotes to avoid possible mistakes, but apparently the linter does not agree with me :D

mendaomn (approved):

PR #291, opened 2022-09-05
Allow update.sh to update multiple stacks

BarcoMasile:
## Description
Up until now the`update.sh` script works correctly only if the account has only one PCM deployed in the specified region.
This PR aims at providing the ability to select a specific stack to update with the latest version of PCM.
This is done through a new mandatory parameter to pass which is `--stack-name STACKNAME`.

## Changes
Providing a stack name allows the script to retrieve the exact name of the lambda function to get the ARN for (belonging to the specified stack), so that the update action doesn't make any assumption of name prefix.
Plus, the private repo to update il also the one belonging the correct stack (specified through the new parameter of the script).

## Changelog entry
Our internal update process now allows (and correctly handles) multiple PCluster Manager deployments in a single account.

## How Has This Been Tested?
Since no change was made to how the update.sh script works, but only how parameters are retrieved, the test performed aimed at checking that with the previous version of the script and with the new one, the same values are correctly retrieved from my personal account.

### DIFF
--- a/scripts/update.sh
+++ b/scripts/update.sh
@@ -5,7 +5,7 @@ REGION_SET=false
 LOCAL=false
 TAG=latest
 
-USAGE="$(basename "$0") [-h] [--region REGION] [--tag TAG] [--local]"
+USAGE="$(basename "$0") [-h] [--stack-name STACKNAME] [--region REGION] [--tag TAG] [--local]"
 
 while [[ $# -gt 0 ]]
 do
@@ -15,6 +15,11 @@ case $key in
     -h)
     echo "$USAGE" >&2
     exit 1
+    ;;
+    --stack-name)
+    STACKNAME="$2"
+    shift # past argument
+    shift # past value
     ;;
     --region)
     REGION=$2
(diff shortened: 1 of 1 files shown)


scripts/update.sh
```diff
@@ -16,6 +16,11 @@ case $key in
     echo "$USAGE" >&2
     exit 1
     ;;
+    --stack-name)
+    STACKNAME="$2"
+    shift # past argument
```
tmscarla, line 21: nitpick: looks like redundant comments to me, but I guess you put them there to be consistent with the other arguments. We can address them in another PR! :)

scripts/update.sh
```diff
@@ -16,6 +16,11 @@ case $key in
     echo "$USAGE" >&2
     exit 1
     ;;
+    --stack-name)
+    STACKNAME="$2"
```
tmscarla, line 20: question: which is the stack name? Since PCM has multiple nested stacks are we talking about the "main" one? I guess so..

tmscarla (approved):
LGTM!

PR #294, opened 2022-09-06
Fix CloudFormation invocation in update.sh script + "bash not found" error

BarcoMasile:
## Description
In the previous PR #291 there was an hardcoded string used, instead of the correct parameter introduced in that PR, plus the GH action presents a new error `bash not found` that we're trying to solve.

## Changes
- github-env-setup.yml has been edited to introduce the newly necessary policy the the role used, in order to perform the DescribeStackResources action
- the package.json `prepare` script now uses `/bin/bash` instead of `bash` only to perform its operations
- the update.sh script correctly uses the `--stack-name` parameter and the `region` value to perform its operation against AWS APIs.

## How Has This Been Tested?
This was tested both locally and on GitHub actions deploying the code and temporarily changing the GH actions to perform scripts only on the wip branch. 

## References
#291 
#292 

## PR Quality Checklist

- [x] I removed hardcoded strings

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -37,7 +37,7 @@
     "lint": "next lint",
     "dev": "next dev",
     "export": "next build && next export -o build",
-    "prepare": "bash ./scripts/git-secrets-command.sh '--register-aws > /dev/null' && cd .. && husky install frontend/.husky",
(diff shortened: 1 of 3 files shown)


tmscarla (approved):
LGTM :) 

nit++: if you past the link of other PRs as plain text GitHub will automatically convert it with a pretty link!

PR #295, opened 2022-09-06
release process - Fix the 1-click links adding also an alternative link

eantonin:
## Description

release process - Fix the 1-click links adding also an alternative link

## Changes

release process - Fix the 1-click links adding also an alternative link

This will remain a draft, can be merged only at release time

The description and all the testing part is the same as https://github.com/aws-samples/pcluster-manager/pull/293


README.md
```diff
@@ -11,35 +11,35 @@ You can get started with your first cluster in as little as 15 minutes using the
...
+| Ohio (us-east-2)  | [![Launch](https://samdengler.github.io/cloudformation-launch-stack-button-svg/images/us-east-2.svg)](https://us-east-2.console.aws.amazon.com/cloudformation/home?region=us-east-2#/stacks/create/review?stackName=pcluster-manager&templateURL=https://pcm-release-us-east-1.s3.us-east-1.amazonaws.com/pcluster-manager.yaml)       | [![Launch](https://samdengler.github.io/cloudformation-launch-stack-button-svg/images/us-east-2.svg)](https://us-east-2.console.aws.amazon.com/cloudformation/home?region=us-east-2#/stacks/create/review?stackName=pcluster-manager&templateURL=https://pcm-release-eu-west-1.s3.eu-west-1.amazonaws.com/pcluster-manager.yaml)       |
```
sean-smith, line 18: This seems like a dangerous precedent, we're moving from a region isolated deployment model to a two region deployment model. What happens if both of these regions are down but a customer wants to deploy in `us-east-2`?

mtfranchetto: This is only for the initial loading of PCM's S3 template. Two regions seem fairly reasonable for the scope - and we can save maintenance time by having fewer buckets - as it's very unlikely to have S3 completely offline in both regions, but I understand your concern. Do you have evidence of a similar thing happened in the past?

mtfranchetto (approved):

PR #297, opened 2022-09-08
Cluster in CREATE_FAILED state cannot be selected/deleted

### LINKED ISSUE #271: Cluster in `CREATE_FAILED` state cannot be selected/deleted
When I try to click on a cluster in `CREATE_FAILED` state, I get the following error message. As I cannot select the cluster,I also cannot delete it. Deleting the cfn stack through the cfn console works fine though.

```
"Error (test1): Unable to parse configuration file. Unable to load configuration from bucket 'parallelcluster-59b87411cc92e432-v1-do-not-delete/parallelcluster/3.2.0/clusters/test1-6eq6qmec6ui5aqkn'. The specified version does not exist."
```

mendaomn:
## Description

Fixes #271 

This PR introduces a check to make sure we don't fetch cluster details when the cluster is in an unstable states. ~~States are listed and can be easily changed in the future~~

Whenever the describe cluster fails, the app shows an error message and stops fetching other data (preventing misleading errors from happening)

If we want to prevent the first error message from showing, we should move the error handling outside of the `DescribeCluster` command and handle it on a case-by-case basis (I'd do this in another PR anyway)

## Changes

- avoid describing the cluster if it is is an unstable state
- fix error message showing `&nbsp;` symbols

## Changelog entry

Fix cluster selection not working when cluster is in CREATE_FAILED state

## How Has This Been Tested?

- unit tests
- locally, mocking the state of the cluster
- mocked a `CREATE_FAILED` and deleted the cluster successfully
- mocked a `CREATE_FAILED` and selected the cluster successfully

## PR Quality Checklist

### DIFF
--- a/frontend/src/old-pages/Clusters/util.tsx
+++ b/frontend/src/old-pages/Clusters/util.tsx
@@ -10,24 +10,28 @@
 // @ts-expect-error TS(7016) FIXME: Could not find a declaration file for module 'js-y... Remove this comment to see the full error message
 import jsyaml from 'js-yaml'
 
-import {DescribeCluster, GetConfiguration} from '../../model'
-import {clearState, getState, setState} from '../../store'
+import {getState, setState} from '../../store'
 import {getIn} from '../../util'
 
-export function selectCluster(clusterName: any, navigate: any) {
+export async function selectCluster(
+  clusterName: string,
+  DescribeCluster: (name: string, callback?: () => void) => Promise<any>,
+  GetConfiguration: (name: string, callback: (value: any) => void) => void,
+) {
   const oldClusterName = getState(['app', 'clusters', 'selected'])
-  const name = clusterName
-  let config_path = ['clusters', 'index', name, 'config']
-  clearState(['clusters', 'index', name])
-  GetConfiguration(name, (configuration: any) => {
-    setState(['clusters', 'index', name, 'configYaml'], configuration)
-    setState(config_path, jsyaml.load(configuration))
-  })
-  DescribeCluster(name, () => {
-    navigate && navigate('/clusters')
-  })
+  let config_path = ['clusters', 'index', clusterName, 'config']
+  try {
+    await DescribeCluster(clusterName)
+    GetConfiguration(clusterName, (configuration: any) => {
+      setState(['clusters', 'index', clusterName, 'configYaml'], configuration)
+      setState(config_path, jsyaml.load(configuration))
+    })
+  } catch (_) {
+    // NOOP
+  }
+
   if (oldClusterName !== clusterName)
-    setState(['app', 'clusters', 'selected'], name)
+    setState(['app', 'clusters', 'selected'], clusterName)
 }
 
 export function getScripts(customActions: Record<string, any> | null) {
--- a/frontend/src/old-pages/Clusters/util.tsx
+++ b/frontend/src/old-pages/Clusters/util.tsx
@@ -20,6 +20,10 @@ export async function selectCluster(
 ) {
   const oldClusterName = getState(['app', 'clusters', 'selected'])
   let config_path = ['clusters', 'index', clusterName, 'config']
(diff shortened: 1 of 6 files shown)


frontend/src/old-pages/Clusters/util.tsx
```diff
@@ -9,25 +9,61 @@
...
 
-import {DescribeCluster, GetConfiguration} from '../../model'
 import {clearState, getState, setState} from '../../store'
+import {ClusterInfoSummary, ClusterStatus} from '../../types/clusters'
 import {getIn} from '../../util'
 
-export function selectCluster(clusterName: any, navigate: any) {
+export function canFetchClusterDetails(cluster?: ClusterInfoSummary) {
+  switch (cluster?.clusterStatus) {
+    case ClusterStatus.CreateInProgress:
```
eantonin, line 20: Question(possibly blocking): not all the states are here (I'm thinking about update_rollback_complete, update_rollback_in_progress and so on) are you sure we should return false in those cases?

frontend/src/old-pages/Clusters/__tests__/util.test.tsx
```diff
@@ -74,3 +85,110 @@ describe('Given a function to get script names of all custom actions', () => {
...
+        expect(mockGetConfiguration).toHaveBeenCalledTimes(0)
+      })
+
+      it('should not describe the cluster', () => {
+        expect(mockDescribeCluster).toHaveBeenCalledTimes(0)
+      })
+    },
+  )
+
+  describe('when describing the cluster fails', () => {
```
eantonin, line 178: question(non-blocking): I don't see a test in which Describe succeeds and GetConfiguration fails, but I guess test surface here is almost zero right?

eantonin (commented):
From the code/the commits/the PR description I do not understand if you managed to delete the cluster from the unstable states, that reading the issue, should be possible

mendaomn:
> From the code/the commits/the PR description I do not understand if you managed to delete the cluster from the unstable states, that reading the issue, should be possible

Good point, I'll expand the "testing" section of the PR

frontend/src/old-pages/Clusters/__tests__/util.test.tsx
mendaomn: At the moment there is no specific action to take in that scenario (it would be handled as any other request failure, meaning a red alert with its error message)

Would you treat that case in a specific way? What should the app when the cluster can be described, but its configuration fetch failed?

frontend/src/old-pages/Clusters/util.tsx
mendaomn: This part is a bit tricky, I want to start by "allowing" only the states that we know make sense. This way we have complete control and there is no risk of unwanted states resulting in a `DescribeCluster`. Here is the complete list of the supported clusterStatuses:

```ts
export enum ClusterStatus {
  CreateInProgress = 'CREATE_IN_PROGRESS',
  CreateFailed = 'CREATE_FAILED',
  CreateComplete = 'CREATE_COMPLETE',
  DeleteInProgress = 'DELETE_IN_PROGRESS',
  DeleteFailed = 'DELETE_FAILED',
  DeleteComplete = 'DELETE_COMPLETE',
  UpdateInProgress = 'UPDATE_IN_PROGRESS',
  UpdateComplete = 'UPDATE_COMPLETE',
  UpdateFailed = 'UPDATE_FAILED',
}
```

Do you see any state here that you would treat differently?
Thanks!

eantonin: Agreed offline to try to change the approach

frontend/src/old-pages/Clusters/__tests__/util.test.tsx
eantonin: I'd consider that a "no test because not enough test surface"

eantonin (approved):
LGTM, great work

PR #304, opened 2022-09-19
Keep track of the OIDC provider

ermanno:
### Notes
This patch introduces two changes in order to keep track of which OIDC provider we are using:
- We add an environmental variable with the OIDC provider the customer selected when deploying the Cfn template.
- We return the value of the variable in the `get_aff_config` function.

### Tests
- Deployed Cfn template.
- Run unit tests.

### PR Quality Checklist


api/PclusterApiHandler.py
```diff
@@ -171,6 +172,7 @@ def get_app_config():
   return {
     "auth_url": AUTH_URL,
     "client_id": CLIENT_ID,
+    "oidc_provider": OIDC_PROVIDER,
```
BarcoMasile, line 175: suggestion(non-blocking): Since we know which OIDC providers we are supporting, couldn't we use [python enums](https://docs.python.org/3/library/enum.html) instead of relying on strings?

Plus, the enum class [already provides an easy way](https://stackoverflow.com/a/41409392/3973689) of obtaining an enum from a string

ermanno: I wouldn't make this change at this point: at the moment, we get this value from the environment variable and the backend doesn't have a list of supported providers, this is the `AllowedValues` of the `OIDCProvider` parameter of the `pcluster-manager.yaml` Cfn template.

BarcoMasile (approved):

PR #307, opened 2022-09-20
Add parameters to support Okta

ermanno:
### Notes
This patch adds parameters to support Okta. Note that it is not possible to dynamically hide or disable a configuration section based on the choosen OIDC provider in CloudFormation, however it is possible to separate the parameters in different sections using the `AWS::CloudFormation::Interface` metadata, which is the approach adopted in this PR.

image

### Tests
Verified that by choosing Cognito everything continues to work correctly.

### PR Quality Checklist

### DIFF
--- a/infrastructure/pcluster-manager.yaml
+++ b/infrastructure/pcluster-manager.yaml
@@ -38,10 +38,32 @@ Parameters:
     Type: String
     Default: ''
   OIDCProvider:
-    Description: OICD provider to use (defaults to Amazon Cognito).
+    Description: OIDC provider to use (defaults to Amazon Cognito).
     Type: String
-    AllowedValues: [Cognito]
+    AllowedValues: [Cognito, Okta]
     Default: Cognito
+  OktaDomain:
+    Description: The Okta domain.
+    Type: String
+    Default: ''
+  OktaResourceServerId:
+    Description: The Okta resource server ID.
+    Type: String
+    Default: ''
+  OktaClientId:
+    Description: The client ID.
+    Type: String
+    Default: ''
+  OktaClientSecret:
+    Description: The client secret.
+    Type: String
+    NoEcho: true
+    Default: ''
+  OktaScopesList:
+    Description: The Okta scopes list.
+    Type: String
+    Default: 'openid profile email groups'
+
 
 Metadata:
   AWS::CloudFormation::Interface:
@@ -61,6 +83,18 @@ Metadata:
         Parameters:
           - Version
           - PublicEcrImageUri
+      - Label:
+          default: OIDC provider
+        Parameters:
+          - OIDCProvider
+      - Label:
+          default: Okta - Compile only if Okta is selected as OIDC provider
+        Parameters:
+          - OktaDomain
+          - OktaResourceServerId
+          - OktaClientId
+          - OktaClientSecret
+          - OktaScopesList
     ParameterLabels:
       EnableMFA:
         default: Require Multi-Factor Authentication for all Users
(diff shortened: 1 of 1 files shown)


infrastructure/pcluster-manager.yaml
```diff
@@ -70,6 +99,16 @@ Metadata:
         default: Initial Admin's Email 
       AdminUserPhone:
         default: Initial Admin's Phone Number
+      OIDCProvider:
+        default: OIDC provider
+      OktaDomain:
+        default: Domain
+      OktaServerId:
+        default: Server Id
```
BarcoMasile, line 107: issue(non-blocking): "domain" and "Server id" are too vague. Since it's the OktaDomain and the authorizationServerId, I would prepend the work okta to the domain, and maybe even to the authorizationServerId (I'll leave the choice to you) so to make it clear what these refers to.

infrastructure/pcluster-manager.yaml
```diff
@@ -61,6 +79,17 @@ Metadata:
...
+          default: OIDC provider
+        Parameters:
+          - OIDCProvider
+      - Label:
+          default: Okta - Compile only if Okta is selected as OIDC provider
+        Parameters:
+          - OktaDomain
+          - OktaServerId
+          - OktaClientSecret
+          - OktaScopesList
```
BarcoMasile, line 92: issue(blockgin): OktaClientId is missing from this list and is necessary

infrastructure/pcluster-manager.yaml
```diff
@@ -38,10 +38,28 @@ Parameters:
...
+    Description: OIDC provider to use (defaults to Amazon Cognito).
     Type: String
-    AllowedValues: [Cognito]
+    AllowedValues: [Cognito, Okta]
     Default: Cognito
+  OktaDomain:
+    Description: The Okta domain.
+    Type: String
+    Default: ''
+  OktaServerId:
```
BarcoMasile, line 49: This should be OktaResourceServerId

infrastructure/pcluster-manager.yaml
```diff
@@ -38,10 +38,28 @@ Parameters:
...
+    Description: The ID of the Okta authorization server.
+    Type: String
+    Default: ''
+  OktaClientSecret:
+    Description: The client secret.
+    Type: String
+    NoEcho: true
+    Default: ''
```
BarcoMasile, line 57: issue(blockgin): as said in the comment below, we also need OktaClientId here (it could be NoEcho: false since it's not super sensible)

BarcoMasile (changes requested):
One other thing is missing, and it's the point in which in the template, we make these values available to the Lambda configuration as Environment variables.

In particular, CLIENT_ID, CLIENT_SECRET and SCOPES_LIST should have these names (so they override the value of the same env vars used for Cognito).
The var about okta domain and the other one about resource server id should have OKTA_ prefix in the env var name

infrastructure/pcluster-manager.yaml
ermanno: To clarify, the actual parameter is called `OktaServerId`, the `Server Id` is the label visible in the CloudFormation console, in this case I have omitted the "Okta" prefix since these parameters appear in the "Okta" section.

mendaomn:
Please try to run `update_infrastructure.sh` on the demo environment before merging

ermanno:
>

ermanno:
The `CLIENT_SECRET` will var will be changed to take the secret name, the idea is to manage it in a similar way as to what we do for Cognito (https://github.com/aws-samples/pcluster-manager/blob/main/infrastructure/pcluster-manager-cognito.yaml#L51-L137).

BarcoMasile (approved):

PR #311, opened 2022-10-03
Add mechanism to support feature flags

mendaomn:
## Description

This PR introduces the components to support feature flags in the application. Another PR will follow up with adoption of such components

## Changes

- add `featureFlagsProvider` to retrieve flags (in the future they may come from the network)
- add `useFeatureFlag` hook to be used in views
- add new devDependency `@testing-library/react-hooks`
- refactored `HeadNode.tsx` to showcase use of new `useFeatureFlag` mechanism

## How Has This Been Tested?

- unit tests
- mocked different versions and checked the `HeadNode` UI change accordingly

## PR Quality Checklist


frontend/src/feature-flags/excludeFeatureRules.ts
```diff
@@ -0,0 +1,12 @@
+import {ExcludeFeatureRule} from './featureFlagsProvider'
+
+export const EXCLUDE_FEATURE_RULES: ExcludeFeatureRule[] = [
```
mtfranchetto, line 3: suggestion: to simplify these checks and also improve the readability, you can keep a map of features for every version, and build the supported features with a reduce.
```
const FEATURES_PER_VERSION = {
  '3.1.0': ['multiuser_cluster'],
  '3.2.0': ['lustre_persistent2', ...],
  '3.3.0': ['feature_x', ...], 
}

const supportedFeatures = FEATURES_PER_VERSION.reduce((acc, currentFeatures, version) => {
  if (userVersion >= version) {
    return [...acc, ...currentFeatures];
  }
  return acc;
}, []);
```

frontend/src/feature-flags/excludeFeatureRules.ts
```diff
@@ -0,0 +1,12 @@
...
+  (feature, minorVersion) => feature === 'fsx_ontap' && minorVersion < 2,
```
mtfranchetto, line 4: issue: this should be `<= 2`, as well as the other features, since it has been introduced with PC 3.2.0

frontend/src/feature-flags/featureFlagsProvider.ts
```diff
@@ -0,0 +1,29 @@
...
+  feature: SupportedFeature,
+  minorVersion: number,
+) {
+  return (
+    exclusionRules.map(rule => rule(feature, minorVersion)).filter(Boolean)
+      .length === 0
+  )
+}
+
+export async function featureFlagsProvider(
```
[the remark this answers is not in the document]
mendaomn: Yep, in the next PR I prepared a `useFeatureFlag(flag: SupportedFeature): boolean` and the corresponding `FeatureManagerContext: Context<IFeatureManager>`

I'm not sure the `featureFlagsProvider` should access the app store itself (even though I realize receiving the `minorVersion` is too specific)

Here's the planned usage of this APIs:

```tsx
// in views
const fsxOnTapEnabled = useFeatureFlag('fsx_ontap') // true | false

// in _app
const minorVersion = useState(['app', 'version', 'minor'])
const activeFeatures = await featureFlagsProviderFunction(
    EXCLUDE_FEATURE_RULES,
    minorVersion,
  )
const featureManager = new FeatureManager()
featureManager.setFeatures(activeFeatures)

// ...
<FeatureManagerContext.Provider value={featureManager}>
   <Component {...pageProps} />
</FeatureManagerContext.Provider>
```

frontend/src/feature-flags/excludeFeatureRules.ts
mendaomn: Currently we are checking that `versionMinor >= 2` to allow this feature, therefore in the exclude rule I believe it's correct to check that it's `< 2`

frontend/src/feature-flags/excludeFeatureRules.ts
mendaomn: I like the idea, I discarded this approach as it's specific to versions while it could make sense that features get disabled according to some different logic in the future (e.g. something about the user)

But I agree it's more readable, therefore we can change it if we don't want to think about other types of flags for now

frontend/src/feature-flags/excludeFeatureRules.ts
mtfranchetto: Yes, you're right. However, it could be better to think on "allowing" rather than "excluding" as it probably makes more sense in the feature flags context.

frontend/src/feature-flags/featureFlagsProvider.ts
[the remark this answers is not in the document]
mendaomn: Refactored and added the hook

frontend/src/feature-flags/excludeFeatureRules.ts
mendaomn: As per our discussion, I refactored into a mechanism based on "allowing" rather than "excluding"

frontend/src/feature-flags/greaterVersionThan.ts
```diff
@@ -0,0 +1,20 @@
...
+// Licensed under the Apache License, Version 2.0 (the "License"). You may not use this file except in compliance
+// with the License. A copy of the License is located at
+//
+// http://aws.amazon.com/apache2.0/
+//
+// or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
+// OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
+// limitations under the License.
+
+export function greaterVersionThan(versionA: string, versionB: string) {
```
mtfranchetto, line 11: suggestion: there's no need to compare semver strings manually. Greater/less than operators work well:
```
'3.2.1' >= '3.2.1'   // true
'3.2.1' > '3.2.1'   // false
```

mtfranchetto (approved):

frontend/src/feature-flags/greaterVersionThan.ts
mendaomn: Great point, will do

mtfranchetto (approved):

PR #317, opened 2022-10-13
Production deploy workflow

mtfranchetto:
## Description

This is the first version of the workflow used to release PCM to our customers.

## Changes

- remove backup logic from release scripts which is not needed since those files are discarded automatically at the end of the run
- new IAM role to deploy to our public ECR registry
- added a new workflow, mimicking what we have done manually to release in production

## How Has This Been Tested?

Deployed to the new buckets/registry using a custom branch. [The triggered run](https://github.com/aws-samples/pcluster-manager/actions/runs/3235521704) has completed successfully, and the released file have been deployed with a 1-click CFN. The product is working correctly.

## PR Quality Checklist

### DIFF
--- a/.github/workflows/production_release.yml
+++ b/.github/workflows/production_release.yml
@@ -1,0 +1,81 @@
+name: Release to production
+
+on:
+  push:
+    #tags: '*'
+    branches: production-deploy
+
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  frontend-tests:
+    runs-on: ubuntu-20.04
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v3
+
+      - uses: actions/setup-node@v3
+        name: Setup Node version
+        with:
+          node-version-file: frontend/.nvmrc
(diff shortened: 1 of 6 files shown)


mendaomn (approved):

.github/workflows/production_release.yml
```diff
@@ -0,0 +1,80 @@
+name: Release to production
+
+on:
+  push:
+    tags: '*'
```
tmscarla, line 5: Will this workflow be triggered for every tag on each branch?

If this is the case, we should state it clearly in the *How to contribute* section, or at least scope it down using a regex, for instance: `^v?(\d+(?:\.\d+)*)$`

tmscarla (approved):

.github/workflows/production_release.yml
tmscarla: I see no threats, just risks on maintainers side in case a tag is pushed and an involuntary deployment is triggered.

Anyway, I am good with it :)

PR #334, opened 2022-10-25
Deploy documentation only on releases

mtfranchetto:
## Description

Changes the trigger of public documentation workflow to deploy updates only when a release is made. This is to avoid inadvertent deployments of documentation regarding unreleased PCM or PC versions.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -1,10 +1,8 @@
-name: github pages
+name: Deploy public documentation
 
 on:
   push:
-    branches:
-      - main  # Set a branch to deploy
-  pull_request:
+    tags: '*'
 
 jobs:
   deploy:
--- a/decisions/20221025-public-documentation-release.md
+++ b/decisions/20221025-public-documentation-release.md
@@ -1,0 +1,13 @@
+# Public documentation release
+
+- Status: accepted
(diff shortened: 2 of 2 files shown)


.github/workflows/gh-pages.yml
```diff
@@ -1,10 +1,8 @@
-name: github pages
+name: Deploy public documentation
 
 on:
   push:
-    branches:
-      - main  # Set a branch to deploy
-  pull_request:
+    tags: '*'
```
lazyoft, line 5: question(non-blocking): do we want to prefix releases so that this will not trigger anytime a tag is pushed in the repository?

lazyoft (approved):
Maybe we can write an ADR based on the implicit decision taken by this PR

.github/workflows/gh-pages.yml
mtfranchetto: It's the same as we do for [production release](https://github.com/aws-samples/pcluster-manager/blob/main/.github/workflows/production_release.yml): since all tags are protected (meaning they can be created or removed only by maintainers) I don't think it's a real issue. GitHub creates automatically a tag when opening a new release, so no one has to create or update a tag by hand.

tmscarla (dismissed):

lazyoft (approved):

PR #352, opened 2022-11-08
Filter out clusters in configuration source 'From Cluster' dropdown in case they do not match up to the minor version of PC

tmscarla:
## Description

Currently it is possible to choose as a template a previously created cluster from the wizard to use for a new cluster the user wants to create.

ParallelCluster doesn't allow cluster creation with configuration versions different from the PC version in use. So using a configuration from a previously created cluster (with a cluster config version from a different PC version) would raise an error on the PC side.

This PR aims to solve this issue, by filtering out clusters in the 'From Cluster' dropdown in case they do not match the exact version of PC.

## How Has This Been Tested?

* Exported locally `API_VERSION` env variable
   * Verified that clusters with the same version are displayed in the dropdown
   * Verified that clusters with a different version are not displayed in the dropdown

## PR Quality Checklist

### DIFF
--- a/frontend/src/old-pages/Configure/Source.tsx
+++ b/frontend/src/old-pages/Configure/Source.tsx
@@ -128,9 +128,17 @@ function sourceValidate(suppressUpload = false) {
   return valid
 }
 
+function checkMinorVersion(v1: string, v2: string) {
+  return v1.split('.', 2).join('.') === v2.split('.', 2).join('.')
+}
+
 function ClusterSelect() {
   const selectedPath = ['app', 'wizard', 'source', 'selectedCluster']
+  const apiVersion = useState(['app', 'version', 'full'])
   const clusters = useState(['clusters', 'list']) || []
+  const filteredClusters = clusters.filter((cluster: ClusterInfoSummary) =>
+    checkMinorVersion(cluster.version, apiVersion),
+  )
   const selected = useState(selectedPath)
   const errors = useState([...sourceErrorsPath, 'sourceClusterName'])
   let source = useState([...sourcePath, 'type'])
@@ -139,7 +147,7 @@ function ClusterSelect() {
   const itemToOption = (item: ClusterInfoSummary) => {
     if (item && item.clusterStatus != ClusterStatus.DeleteInProgress)
       return {label: item.clusterName, value: item.clusterName}
-    else return {label: i18next.t('wizard.source.clusterSelect.placeholder')}
+    else return null
   }
 
   return (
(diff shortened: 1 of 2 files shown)


frontend/src/old-pages/Configure/Source.tsx
```diff
@@ -130,7 +130,11 @@ function sourceValidate(suppressUpload = false) {
 
 function ClusterSelect() {
   const selectedPath = ['app', 'wizard', 'source', 'selectedCluster']
+  const apiVersion = useState(['app', 'version', 'full'])
   const clusters = useState(['clusters', 'list']) || []
+  const filteredClusters = clusters.filter(
+    (cluster: ClusterInfoSummary) => cluster.version === apiVersion,
```
mtfranchetto, line 136: question: are we sure we want to do an exact version check here? Only minor versions of PC introduce new features and fields, so we can be probably safe allowing an import between patch versions.
For example, if I'm running PC 3.2.1 and I want to import a cluster to from 3.2.0, PCM should allow that.

tmscarla: Agree, we should be safe. Going to check up to the minor only.

lazyoft: I'm not sure about this for the time being, because whilst it's true that a patch version should contain only bug fixes we still don't have a clear definition of what a patch version might be. For the time being I would keep it safe and maybe widen the scope with a subsequent release, when we will have an agreement of what a patch version is.

frontend/src/old-pages/Configure/Source.tsx
```diff
@@ -128,9 +128,17 @@ function sourceValidate(suppressUpload = false) {
   return valid
 }
 
+function checkMinorVersion(v1: string, v2: string) {
+  return v1.split('.', 2).join('.') === v2.split('.', 2).join('.')
```
mtfranchetto, line 132: praise: great to see you've used the `limit` parameter in `split`, clever move!

frontend/src/old-pages/Configure/Source.tsx
tmscarla: Shouldn't be intrinsic what a patch version is in semver? Or PC is not adhering completely to the specific?

mtfranchetto (approved):

frontend/src/old-pages/Configure/Source.tsx
mtfranchetto: SemVer is followed correctly if you take a look at the [release history](https://docs.aws.amazon.com/parallelcluster/latest/ug/document_history.html)

PR #360, opened 2022-11-10
Use explicit versioning as tagging strategy for releases

BarcoMasile:
## Description
When performing a release, the latest image got tagged as "latest" and as date formatted string.
Now the date string strategy has been replaced with explicit versioning.

## Changes
- edited `pcluster-manager.yml` to target explicitly release 3.3.0
- added a pcm.version file containing the version to use to tag the lambda image (to decouple it from the script)
- build_awslambda.sh with the updated strategy
- removed two variables from the top of the file that were clearly leftover and caused confusion in my head everytime

## How Has This Been Tested?
Yes, manually executing the script in my personal public repo.

After this PR I will also tag the latest image available as 3.3.0.


mendaomn:
Couple of observations:

- we should probably update docs on how to bump the PCM version, since now we need to change the default ECR URI value as well

- > After this PR I will also tag the latest image available as 3.3.0.

This needs to happen before the merge, imho

pcm.version
```diff
@@ -0,0 +1 @@
+3.3.0
```
mtfranchetto, line 1: suggestion: what about having this version in the `package.json` of the frontend?
It can be retrieved at release time with `npm pkg get version` and we avoid having a new file just for versioning

mtfranchetto:
> After this PR I will also tag the latest image available as 3.3.0.

Why? Is it for "completeness"? This code does not get released so it should not raise issues

pcm.version
BarcoMasile: This is a great idea!! Thanks
It'll be in the next revision!

mendaomn:
> > After this PR I will also tag the latest image available as 3.3.0.
> 
> Why? Is it for "completeness"? This code does not get released so it should not raise issues

Yeah, just a minor obsession of mine. It is true it doesn't "need" to happen before the merge

BarcoMasile:
> > After this PR I will also tag the latest image available as 3.3.0.
> 
> Why? Is it for "completeness"? This code does not get released so it should not raise issues

It is not needed, but I would do it for completeness and to have an "aligned" situation, since it's easy and quick to do

mtfranchetto (approved):

PR #362, opened 2022-11-11
Introduce logger for Python backend

tmscarla:
## Description

In order to start collecting logs to ease troubleshooting, this PR aims at introducing a `DefaultLogger` class, which leverages the standard Python `logging` library. If running locally, log level is set to `DEBUG`, `INFO` otherwise.

Since Python logger in AWS Lambda has a preset format, we set a custom format if running locally only.

## How Has This Been Tested?
* Added dummy log to list_users() endpoint
   * Verified locally that when the endpoint is hit, the logger logs to console as expected, following the provided format
   * Updated the Lambda image in my personal stack and verified that 

## PR Quality Checklist


api/logger.py
```diff
@@ -0,0 +1,35 @@
...
+  def info(self, msg, extra={}):
+    self.logger.info(self._log_output(msg, extra))  
+
+  def warning(self, msg, extra={}):
+    self.logger.warning(self._log_output(msg, extra))   
+    
+  def error(self, msg, extra={}):
+    self.logger.error(self._log_output(msg, extra))
+
+  def critical(self, msg, extra={}):
```
BarcoMasile, line 32: issue(blocking): the default parameter `extra` is not initialized for every call, so it will hold consecutive values added if it's not cleared after every use, you can read more [here](https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python)

api/logger.py
```diff
@@ -0,0 +1,35 @@
...
+class DefaultLogger(object):
+  def __init__(self, is_running_local):
+    self.logger = logging.getLogger("pcluster-manager")
+    if is_running_local:
+      handler = logging.StreamHandler()
+      handler.setFormatter(logging.Formatter("%(asctime)s - %(levelname)s - %(message)s"))
+      self.logger.addHandler(handler)
+      self.logger.setLevel(logging.DEBUG)
+    else:
+      self.logger.setLevel(logging.INFO)
```
BarcoMasile, line 14: issue(non-blocking): if no formatter is defined for prod, python would default to this: `"%(levelname)s:%(name)s:%(message)s` (as you can see [here](https://hg.python.org/cpython/file/5c4ca109af1c/Lib/logging/__init__.py#l1640))
Is this the desired formatter for prod deployment?

tmscarla: Python logger in AWS Lambda has a preset format, which has the following structure:

`%(asctime)s - @requestid - %(level)s - %(message)s `

The available options were:
* Keep this format as is, avoiding to add another handler
* On logger configuration, remove all the existing handlers and attach our custom handler

api/logger.py
tmscarla: Going to follow the advice and set `None` as default

api/logger.py
tmscarla: Since we have the `extra` parameter, does it make sense to leverage it?

For instance we will have something like this:

```
logger.info("Message", extra={"source": "backend"})

File: api/logger.py [hunk 3]
```diff
@@ -0,0 +1,36 @@
+import logging
+from api.utils import running_local
+
+
+class DefaultLogger(object):
```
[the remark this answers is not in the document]
tmscarla: I went with the easiest solution for the time being, by checking the level of the internal implementation of the logger from the standard logging library.

An alternative solution could be define a `setLevel` method, as suggested by @mtfranchetto, and check that it has been called with the expected value.

Finally, we could also check that the standard output displays the value that we log. Let me know what you think about it.

api/logger.py
BarcoMasile: In order to use that `extra` param you would need either a custom formatter or a json formatter to configure the logging module with.
You could use it if you know that certain parameters will always be provided and include it in the format string like, for example, if you know you always have a "client_ip" extra parameter, you could log it including
`%(client_ip)s` in your format string.

tmscarla: The formatter is configured in the following way:

```
"%(asctime)s - %(levelname)s - %(message)s
```

Where `message` is a JSON which can be queried with CloudWatch Insights or parsed in any other way.

PR #365, opened 2022-11-14
Added global exception handler to avoid data leakage from unhandled exceptions in the backend

BarcoMasile:
## Description
Added global exception handler to avoid data leakage from unhandled exceptions in the backend

Important: right now the implementation uses a logger within `api.exception.handlers` module, this WILL NEED to be replaced once #362 is ready and merged.

## Changes
- Added exceptions package `api/exception/__init__.py` with ExceptionHandler to config the Flask app
- Added global exception handler for Boto core errors and Boto ClientErrors
- Added global exception handler for unhandled Exception
- Correctly formatted `api/utils.py` file and added the ExceptionHandler to the `build_flask_app` function

## How Has This Been Tested?
This has been tested manually and with unit tests

## PR Quality Checklist


api/utils.py
```diff
@@ -56,39 +58,47 @@ def to_iso_timestr(time_in: datetime.datetime) -> str:
...
-  if running_local():
-    app = Flask(name)
-    CORS(app)
+    if running_local():
+        app = Flask(name)
+        CORS(app)
+    else:
+        app = Flask(name, static_url_path="", static_folder="frontend/public")
+
+    ExceptionHandler(app, running_local=running_local())
```
mendaomn, line 95: suggestion: This doesn't need to be a class, since you are not actually managing internal state or keeping an instance around

I think you can just export an `init_expection_handlers()`  or simply [decorate the handlers in place](https://flask.palletsprojects.com/en/2.1.x/errorhandling/#registering)

api/exception/tests/conftest.py
```diff
@@ -0,0 +1,26 @@
+import pytest
```
mendaomn, line 1: question: why are we not using the existing `test` folder?

api/utils.py
```diff
@@ -56,39 +58,47 @@ def to_iso_timestr(time_in: datetime.datetime) -> str:
...
 
+
 def disable_auth():
     return os.getenv("ENABLE_AUTH") == "false"
 
+
 def proxy_to(to_url):
-  """
+    """
   Proxies Flask requests to the provided to_url
```
tmscarla, line 72: nitpick: wrong indentation

BarcoMasile: Thanks! I will fix it asap

api/exception/tests/conftest.py
BarcoMasile: I wanted to separate tests by package, but I realize it's "too soon" since I tracked this in our board.
I will move it under the common tests folder, thanks!

api/utils.py
BarcoMasile: Yes but this is more inline with Flask extensions and I'd like to keep this responsibility over there since it's cleaner and it'll help us handle exception handlers better (right now they're not 100% comprehensive)

api/exception/tests/conftest.py
BarcoMasile: I moved the tests under the common folder

mendaomn (approved):

PR #369, opened 2022-11-18
Added public endpoint to push log entry to from the frontend

BarcoMasile:
## Description
Added a public endpoint to allow the frontend to push log entries to the backend.

## Changes
- Added a new exception handler to handle ValueError in case of API validation error
- Added a `/push-log` endpoint to log the log entries from the frontend
- Added tests for newly added ValueError exception handler and far all validation cases of the new endpoint

## How Has This Been Tested?
Unit tests and manual integration tests

## PR Quality Checklist

### DIFF
--- a/api/exception/handlers.py
+++ b/api/exception/handlers.py
@@ -15,6 +15,11 @@ def websocket_mismatch_nop_handler(err):
     """ Handles websocket error caused by HMR in local development"""
     return {}, 200
 
+def value_error_handler(err):
+    descr, code = str(err), 400
+    logger.error(descr, extra=dict(status=code, exception=type(err)))
+    return __handler_response(code, descr)
+
 def global_exception_handler(err):
     try:
         code = err.code
@@ -23,5 +28,9 @@ def global_exception_handler(err):
     descr = str(err)
 
     logger.error(descr, extra=dict(status=code, exception=type(err)))
-    response = {'Code': code, 'Message': 'Something went wrong'}
+    return __handler_response(code)
+
+
+def __handler_response(code, description='Something went wrong'):
+    response = {'Code': code, 'Message': description}
     return jsonify(error=response), code
--- a/api/tests/test_push_log.py
+++ b/api/tests/test_push_log.py
@@ -34,7 +34,7 @@ def test_push_log_controller_with_invalid_message_key(client):
     assert response.json == {
         'error': {
             'Code': 400,
-            'Message': 'Request body missing on or more mandatory fields ["message", "level"]'
+            'Message': 'Request body missing one or more mandatory fields ["message", "level"]'
         }
     }
 
@@ -48,7 +48,7 @@ def test_push_log_controller_with_invalid_level_key(client):
     assert response.json == {
         'error': {
             'Code': 400,
-            'Message': 'Request body missing on or more mandatory fields ["message", "level"]'
+            'Message': 'Request body missing one or more mandatory fields ["message", "level"]'
         }
     }
 
--- a/api/tests/test_push_log.py
+++ b/api/tests/test_push_log.py
@@ -1,4 +1,7 @@
 
+def trim_log(caplog):
+    return caplog.text.replace('\n', '')
+
 def test_push_log_controller_with_valid_json_no_extra(client, caplog):
     request_body = {'message': 'sample-message', 'level': 'info'}
     expected_log = "INFO     pcluster-manager:logger.py:24 {'message': 'sample-message'}"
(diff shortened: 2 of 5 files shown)


api/tests/test_push_log.py
```diff
@@ -0,0 +1,81 @@
+
+def trim_log(caplog):
+    return caplog.text.replace('\n', '')
+
+def test_push_log_controller_with_valid_json_no_extra(client, caplog):
```
tmscarla, line 5: praise: tests look good to me! It seems to me that we have all cases covered properly. +1 for caplog, didn't know about this fixture!

tmscarla (changes requested):
Correct the typo, otherwise LGTM!

api/tests/test_push_log.py
BarcoMasile: Thank you very much!

app.py
```diff
@@ -182,6 +182,25 @@ def login_():
     def logout_():
         return logout()
 
+    @app.route('/push-log', methods=['POST'])
+    def push_log():
+        if 'level' not in request.json or 'message' not in request.json:
+            raise ValueError('Request body missing on or more mandatory fields ["message", "level"]')
```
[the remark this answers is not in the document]
BarcoMasile: Corrected in next revision, thanks!

app.py
```diff
@@ -182,6 +182,25 @@ def login_():
...
+
+        _json = request.json.get
+        level, message, extra = _json('level'), _json('message'), _json('extra')
+        if not (extra is None or type(extra) is dict):
+            raise ValueError('Extra param must be a valid json object')
+
+        logging_fun = getattr(logger, level.lower(), None)
+
+        if logging_fun is None:
+            raise ValueError('Invalid logging level requested')
```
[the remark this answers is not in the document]
BarcoMasile: So I thought about it but I preferred to not use a default level since if by mistake a level was not specified for a "debug" message, then it would end up in the log with info that should probably not be seen

app.py
```diff
@@ -182,6 +182,25 @@ def login_():
...
+
+        logging_fun(message, extra=extra)
```
mendaomn, line 200: issue: I believe the logging endpoint should set the `source` attribute. I don't think we'll have different sources hitting this very endpoint aside from our frontend, and in that case we can override the attribute

Long story short, I'd like logs coming from the FE to be marked with a specific source as part of the log structure (I think you can just add it to the `extra` dict, if I remember correctly @tmscarla )

app.py
```diff
@@ -182,6 +182,25 @@ def login_():
...
 
+    @app.route('/push-log', methods=['POST'])
+    def push_log():
+        if 'level' not in request.json or 'message' not in request.json:
+            raise ValueError('Request body missing one or more mandatory fields ["message", "level"]')
+
+        _json = request.json.get
+        level, message, extra = _json('level'), _json('message'), _json('extra')
+        if not (extra is None or type(extra) is dict):
+            raise ValueError('Extra param must be a valid json object')
```
mendaomn, line 193: question: This looks good to me for the time being, however: how you looked at done-for-you options to validate endpoint inputs?

api/exception/handlers.py
```diff
@@ -23,5 +28,9 @@ def global_exception_handler(err):
     descr = str(err)
 
     logger.error(descr, extra=dict(status=code, exception=type(err)))
-    response = {'Code': code, 'Message': 'Something went wrong'}
+    return __handler_response(code)
+
+
+def __handler_response(code, description='Something went wrong'):
```
mendaomn, line 34: I think you can leverage this also in `boto3_exception_handler`

app.py
BarcoMasile: I've done some research, Flask doesn't provide you with anything unfortunately, but I'm working on an idea on how to add input validation in a uniform manner
I hope to be able to bring this in the near future

api/exception/handlers.py
BarcoMasile: The boto3 exception already has a dict so this `__handler_response` is a way to uniform all errors returned to the nice way boto3 have of doing it

app.py
```diff
@@ -182,6 +182,25 @@ def login_():
     def logout_():
         return logout()
 
+    @app.route('/push-log', methods=['POST'])
```
mendaomn, line 185: issue: shall we protect this endpoint against public access?

BarcoMasile: I thought about this but if we protect it we can't log issues risen before and while the auth process is performed.
We could consider it a tradeoff, I'm not totally sure how to go about this

app.py
BarcoMasile: Done in new review, thanks!

app.py
BarcoMasile: Endpoint is now private in latest revision

mendaomn (approved):

PR #370, opened 2022-11-23
Added frontend logger implementation to push logs to CloudWatch through the backend

BarcoMasile:
## Description
After #369 we now introduce a logger implementation on the frontend side that leverages the /logs endpoint implemented in the previous PR.
It is possible to use the newly introduced logger to push log entries to the backend in async, so they end up in CloudWatch lambda log streams.

## How Has This Been Tested?
- unit tests
- manual tests checking the backend gets the message and logs correctly

## PR Quality Checklist


frontend/src/logger/pcm.logger.ts
```diff
@@ -0,0 +1,75 @@
...
+  critical = 'CRITICAL',
+}
+
+interface LogEntry {
+  message: string
+  level: LogLevel
+  extra?: {[key: string]: string}
+}
+
+export class PcmLogger {
```
mendaomn, line 20: suggestion: I wouldn't use "PCM" in the namings, it's just a Logger pretty much as "Home" is not called "PCMHome"

frontend/src/logger/pcm.logger.ts
```diff
@@ -0,0 +1,75 @@
...
+  warning = 'WARNING',
+  debug = 'DEBUG',
+  error = 'ERROR',
+  critical = 'CRITICAL',
+}
+
+interface LogEntry {
+  message: string
+  level: LogLevel
+  extra?: {[key: string]: string}
```
mendaomn, line 17: issue(non-blocking): if we plan to support nested objects in the `extra` parameter, a better type signature could be `Record<string, unknown>`

frontend/src/logger/pcm.logger.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  private buildMessage(
+    level: LogLevel,
+    message: string,
+    extra?: {[p: string]: string},
+  ): LogEntry {
+    const entry: LogEntry = {message, level}
+    if (extra) {
+      entry['extra'] = extra
+    }
```
mendaomn, line 71: I don't think this is needed, you can safely `return {message, level, extra}`

frontend/src/model.tsx
```diff
@@ -33,6 +34,8 @@ const axiosInstance = axios.create({
   baseURL: getHost(),
 })
 
+const logger = new PcmLogger(axiosInstance, getState(['app', 'appConfig']))
```
mendaomn, line 37: issue: The logger would most likely be used throughout the app, also it depends on the value present in the store. By initializing it here, you'd never update the value of `appConfig` once it reached the store (it depends on an API call which is fired only after this line is executed)

I would suggest that we create a `LoggerContext` to pass the logger down to components. This would allow to react to changes in the store (e.g. the appConfig) and change the logger instance (if needed)

It's also easier to inject in component tests, as you'd simply wrap the component under test with a mocked `LoggerProvider`

In case we needed to reach out for the logger in places where contexts are not available (e.g. pure business logic), we can typically retrieve the logger from the context and pass it to such pure functions (or setup a service container to mimic an IoC container, but I'd wait for the actual need to occur)

frontend/src/logger/__tests__/pcm.logger.test.ts
```diff
@@ -0,0 +1,93 @@
...
+        expect(mockedAxios.post).toHaveBeenCalledTimes(1)
+        expect(mockedAxios.post).toHaveBeenCalledWith(
+          expectedPath,
+          expectedLogEntry,
+          expectedConfig,
+        )
+        expect(response.status).toBe(200)
+      })
+
+      it('logs an error to stderr when incorrect data is passed', async () => {
```
mendaomn, line 73: suggestion: I would argue the effect we'd rather test is the return value of the function, in case of error. I'm not sure we care that much fo the warning being logged in console

frontend/src/logger/pcm.logger.ts
mendaomn, line 20: question: Have you thought of implementing the same interface (e.g. `ILogger`) with a `ConsoleLogger` to be used in local development? Not sure about this one, I'd leave it up to you to consider whether this is needed

frontend/src/logger/pcm.logger.ts
```diff
@@ -0,0 +1,75 @@
...
+
+enum LogLevel {
+  info = 'INFO',
+  warning = 'WARNING',
+  debug = 'DEBUG',
+  error = 'ERROR',
+  critical = 'CRITICAL',
+}
+
+interface LogEntry {
```
mendaomn, line 14: suggestion: I would specify both Request and Response types. Something along the lines of:

```ts
interface PostLogsRequest {}
interface PostLogsResponse {}
```

you can then use the `LogEntry` type to detail the request if you wish, and most importantly, use the PostLogsResponse type in your `.catch`

frontend/src/logger/pcm.logger.ts
```diff
@@ -0,0 +1,75 @@
...
+      ? handleNotAuthorizedErrors(appConfig)
+      : identityFn<Promise<any>>
+  }
+
+  log(logLevel: LogLevel, message: string, extra?: {[key: string]: string}) {
+    const logEntry = this.buildMessage(logLevel, message, extra)
+    return this.handle401and403(
+      this.post('/logs', logEntry, {
+        headers: {'Content-Type': 'application/json'},
+      }),
```
mendaomn, line 36: Sadly this logic is repeated in our `request` function. I guess you did not use the `request` function here since it also adds the region to the URL. I would consider to avoid spreading configuration logic and error handling around. Can be done in another PR though

frontend/src/logger/pcm.logger.ts
BarcoMasile: Yes I thought about it, but since we always run the backend also in dev mode, logs reach the backend and can be seen there, I don't think it would bring much value, but rather making logs easier to see
I'm open to it!

mendaomn (approved):

mendaomn (approved):

PR #379, opened 2022-11-30
Clean up codebase from guest/users roles

mendaomn:
## Description

Following #375, this PR cleans up the codebase removing references to deprecated user and guest roles

## Changes

- remove distinction between `user` and `admin` from the SideBar
- remove `guest` from the authorized user groups
- fail `get_identity` when no groups for the user are set

## How Has This Been Tested?

- unit tests
- manually by following these steps:
    - create a new user via PCM
    - logged in with the new user
    - created a new cluster successfully

## References

- #375 

## PR Quality Checklist

### DIFF
--- a/app.py
+++ b/app.py
@@ -110,7 +110,7 @@ def run():
         return get_dcv_session()
 
     @app.route("/manager/get_identity")
-    @authenticated({"guest"})
+    @authenticated(ADMINS_GROUP)
     def get_identity_():
         return get_identity()
 
--- a/frontend/src/components/SideBar.tsx
+++ b/frontend/src/components/SideBar.tsx
@@ -8,12 +8,15 @@
 // or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 // OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 // limitations under the License.
-import {SideNavigation, SideNavigationProps} from '@cloudscape-design/components'
+import {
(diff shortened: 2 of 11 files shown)


app.py
```diff
@@ -110,7 +110,7 @@ def get_dcv_session_():
         return get_dcv_session()
 
     @app.route("/manager/get_identity")
-    @authenticated({"guest"})
+    @authenticated(ADMINS_GROUP)
```
mtfranchetto, line 113: suggestion(non-blocking): since almost all routes now are accessible only by admin users, it would be nice to introduce authentication by default on all endpoints, and allow access from unregistered users only on authentication methods

frontend/src/components/SideBar.tsx
```diff
@@ -35,17 +38,17 @@ export default function SideBar() {
     React.useMemo(() => {
       return [
         {type: 'link', text: t('global.menu.home'), href: '/home'},
-        isUser()
+        isAdmin()
```
mtfranchetto, line 41: nitpick: I guess we can remove entirely these ternaries since all users are administrators, even in other places of the portal

mtfranchetto (changes requested):

app.py
mendaomn: Agree. Let's do it on a separate PR, so that we can focus on properly test such change

frontend/src/components/SideBar.tsx
mendaomn: Absolutely. Working on it

mtfranchetto (approved):

PR #383, opened 2022-12-01
Added security headers to improve application security posture

BarcoMasile:
## Description
This PR introduces an `after_request` hook to register security headers for every http response returned by PCM to improve the overall security posture.

## Changes
A new flask extension has been introduced to register an after request function that manipulates every `Response` object to add security headers.
In development mode CORS is enabled so headers follow suite.

## How Has This Been Tested?
This has been tested manually in a local environment and in a cloud environment in my personal account

## PR Quality Checklist


api/tests/test_add_response_headers.py
```diff
@@ -0,0 +1,25 @@
+from api.security.headers import SECURITY_HEADERS, CORS_HEADERS
```
mendaomn, line 1: issue: As tedious as it might seem, we shouldn't test against the actual configuration object used by the software under test

This is usually avoided because you'd be replicating the same logic that you are testing, nullifying the test.

Practical example:
- I go in and remove an important header from the config under test
- The test is green and nobody notices

One solution would be enumerating the headers in the test as well

api/tests/test_add_response_headers.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_response_security_headers_dev(dev_app, dev_client):
+    """
+    Given a PCM app in local development
+      When a request is performed
+        Then it should have security headers correctly set for local development
+    """
+    response = dev_client.get('/manager/get_app_config')
```
mendaomn, line 10: I love the idea of testing it by hitting the actual endpoint, however this only test one endpoint. I understand it implicitly tests the app extension as well, but I'd be more explicit and directly test either the `SecurityHeaders` class or the two functions (`add_security_headers_`) to make sure they add the full list of header to a given Response object

It's a bit of a small detail, I'll leave it up to you to decide whether to change this or not

api/tests/test_add_response_headers.py
BarcoMasile: Absolutely right, done in next revision! Thank you!

api/security/headers.py
```diff
@@ -0,0 +1,27 @@
...
+CORS_HEADERS = [
+    {'key': 'Cross-Origin-Resource-Policy', 'default': 'same-site'},
+    {'key': 'Cross-Origin-Embedder-Policy', 'default': 'require-corp'}
+]
+
+SECURITY_HEADERS = [
+    {'key': 'X-Frame-Options', 'default': 'DENY'},
+    {'key': 'X-Content-Type-Options', 'default': 'nosniff'},
+    {'key': 'Referrer-Policy', 'default': 'strict-origin-when-cross-origin'},
+    {'key': 'Strict-Transport-Security', 'default': 'max-age=63072000; includeSubDomains; preload'},
```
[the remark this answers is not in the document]
BarcoMasile: Apparently it does not, I tested manually both in local env and in my personal account (so prod), everything works.
The explanation seems to be that this header only has effects if the website is already in https (which of course is not in local dev). 
But I'm not 100% sure this explanation is factual

mendaomn (approved):

api/tests/test_add_response_headers.py
BarcoMasile: Done in this latest review, I believe your suggestion is a great value so I implemented it
Thanks!

api/security/headers.py
```diff
@@ -0,0 +1,27 @@
+from flask import Response
+
+CORS_HEADERS = [
```
mtfranchetto, line 3: typo: this is `CORP`, not `CORS`

api/utils.py
```diff
@@ -93,6 +94,7 @@ def build_flask_app(name):
         app = Flask(name, static_url_path="", static_folder="frontend/public")
 
     ExceptionHandler(app, running_local=running_local())
+    SecurityHeaders(app, running_local=running_local())
```
mtfranchetto, line 97: suggestion: can the `CORS` management moved inside `SecurityHeaders`? There's already a check for running local, and by doing that we'll end up having all security related stuff in the same place

api/security/headers.py
BarcoMasile: Nice catch, thanks!

api/utils.py
BarcoMasile: I like the idea, I've done it in the next rev coming up, but I had to keep a check. for the different strategy in building the flask app

mtfranchetto (dismissed):

api/utils.py
```diff
@@ -87,14 +87,16 @@ def proxy_to(to_url):
...
-    if running_local():
-        app = Flask(name)
-        CORS(app)
-    else:
-        app = Flask(name, static_url_path="", static_folder="frontend/public")
+    is_running_local = running_local()
+
+    additional_args = {}
+    if is_running_local:
+        additional_args = dict(static_url_path="", static_folder="frontend/public")
```
mendaomn, line 94: issue(blocking): careful, the additional args belong to the "non-local" branch, if I understand correctly

api/security/__init__.py
```diff
@@ -12,6 +13,7 @@ def __init__(self, app: Scaffold = None, running_local=False):
 
     def init_app(self, app: Scaffold):
         if self.running_local:
+            CORS(app)
```
mendaomn, line 16: suggestion: I think this refactor affect the tests of both the SecurityHeader extension and the build_flask_app function

api/utils.py
BarcoMasile: Thank you, fixed in the next rev!

mendaomn (approved):

PR #385, opened 2022-12-05
Enable access logs on the API Gateway

mtfranchetto:
## Description

Enable logging of every call made to the API gateway.

**N.B. the demo environment must be recreated from scratch as soon as this PR gets merged.**

## Changes

- **breaking**: manage manually the stage for the API Gateway so that log settings can be added (it's a breaking change only for stack updates since CloudFormation fails doing that operation)
- add logs settings to the API Gateway stage
- replaced the route matching in place with `$default` as it does exactly what we need to do (match all routes and every method)

## How Has This Been Tested?

Create a new stack and do some manual calls, veryfing the output in the CloudWatch console.

Screenshot 2022-12-02 at 18 16 58
Screenshot 2022-12-02 at 18 17 10

## PR Quality Checklist

### DIFF
--- a/infrastructure/pcluster-manager.yaml
+++ b/infrastructure/pcluster-manager.yaml
@@ -174,16 +174,12 @@ Resources:
       Name: PclusterManager
       Description: PclusterManager Lambda Proxy
       ProtocolType: HTTP
-      Target: !Sub
-        - arn:${AWS::Partition}:apigateway:${AWS::Region}:lambda:path/2015-03-31/functions/arn:${AWS::Partition}:lambda:${AWS::Region}:${AWS::AccountId}:function:PclusterManagerFunction-${StackIdSuffix}/invocations
-        - { StackIdSuffix: !Select [2, !Split ['/', !Ref 'AWS::StackId']] }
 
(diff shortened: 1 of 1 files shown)


mendaomn (approved):

infrastructure/pcluster-manager.yaml
```diff
@@ -200,6 +196,21 @@ Resources:
       PayloadFormatVersion: 2.0
       TimeoutInMillis: 30000
 
+  PCMApiGatewayAccessLog:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      RetentionInDays: 90
```
BarcoMasile, line 202: suggestion: we could specify a log group name so that it's not totally random and autogenerated by CloudFormation

BarcoMasile (approved):

infrastructure/pcluster-manager.yaml
BarcoMasile: I'm ok with it, from CloudFormation documentation it was not clear that the outcome would be like that!
Thanks

PR #386, opened 2022-12-05
Add CSP security header

tmscarla:
## Description

Add `Content-Security-Policy` header to the list of security headers included in every http response returned by PCM.

## Context

Following [CloudScape CSP configuration](https://cloudscape.design/get-started/dev-guides/csp/) and internal recommendations, this is the proposed Content Security Policy for PCM:

```
default-src 'self'; style-src 'self' 'unsafe-inline'; font-src data:; img-src 'self' data:; child-src blob:; object-src 'none'; frame-ancestors 'none'; base-uri 'none';
```

**Directives explanation:**
* **default-src:** fallback for other directives. If a directory is absent, the user agent looks for the default-src directive
* font-src:** specify valid sources for fonts loaded. In CloudScape font-src is needed to ensure that fonts are loaded along with the  rest of the CSS files, to avoid [font display issues](https://css-tricks.com/font-display-masses/).
* **object-src:** valid sources for the [<object>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/object), [<embed>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/embed), and [<applet>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/applet) elements. This elements are considered legacy HTML and should be not used.
* **frame-ancestors:** specify what parent source may embed a page. This directive does not fallback to default-src.
* **base-uri:** restricts the URLs which can be used in a document's [<base>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/base) element

Policy have been evaluated through the CSP evaluator tool as it is shown in the attached image.
Medium severity finding regarding 'self' in default-src → allowed by Aristotle
Recommendation for [require-trusted-types-for](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/require-trusted-types-for) → still an experimental technology which is not supported by all browser as today

CSP header will be included if not running locally only, since it breaks Next HMR while running in development mode. For further details see:
* https://github.com/vercel/next.js/issues/256
* https://github.com/vercel/next.js/issues/14221

## Changelog entry

* Added `Content-Security-Policy` header to the list of security headers included in every http response returned by PCM

## How Has This Been Tested?

* Run `scripts/update.sh` script to update Lambda in my personal test environment
* Verified that no CSP related errors are displayed in console
* Verified that CSP headers are included in each response

## PR Quality Checklist


api/tests/test_add_response_headers.py
```diff
@@ -11,6 +11,8 @@
     'X-XSS-Protection': '1; mode=block'
 }
 
+CSP_HEADER = {'Content-Security-Policy': "default-src 'self'; style-src 'self' 'unsafe-inline'; font-src data:; img-src 'self' data:; child-src blob:; object-src 'none'; frame-ancestors 'none'; base-uri 'none';"}
+
```
BarcoMasile, line 15: nitpick(non-blocking): every time a dev formats his/her code correctly, a puppy smiles

BarcoMasile (approved):

BarcoMasile (dismissed):

api/tests/test_add_response_headers.py
```diff
@@ -11,6 +11,11 @@
     'X-XSS-Protection': '1; mode=block'
 }
 
+CSP_HEADER = {
+    'key': 'Content-Security-Policy',
+    'default': "default-src 'self'; style-src 'self' 'unsafe-inline'; font-src data:; img-src 'self' data:; child-src blob:; object-src 'none'; frame-ancestors 'none'; base-uri 'none';"
+}
+
```
BarcoMasile, line 18: Here it should be:
```python
{
  'Content-Security-Policy': "default-src 'self'; style-src 'self' 'unsafe-inline'; font-src data:; img-src 'self' data:; child-src blob:; object-src 'none'; frame-ancestors 'none'; base-uri 'none';"
}
```

BarcoMasile (approved):

api/tests/test_add_response_headers.py
tmscarla: Will indent properly in next revision!

PR #391, opened 2022-12-06
SSH Command

sean-smith:

## Description

image

image

Users can easily get to the CloudFormation stack from the Stack Events tab. This PR removes the ugly link and adds in a new SSH command.

## Changes

* add SSH command
* remove CFN arn

## Changelog entry

* Include SSH Command in Cluster Summary tab

## PR Quality Checklist

- [x] ~I added tests to new or existing code~

### DIFF
--- a/frontend/src/old-pages/Clusters/Properties.tsx
+++ b/frontend/src/old-pages/Clusters/Properties.tsx
@@ -10,7 +10,7 @@
 // limitations under the License.
 import {ClusterStatus, ClusterDescription} from '../../types/clusters'
 import React from 'react'
-
+import {useTranslation} from 'react-i18next'
 import {findFirst, clusterDefaultUser} from '../../util'
 import {
   getState,
@@ -54,6 +54,7 @@ const ValueWithLabel = ({label, children}: any) => (
 )
 
 export default function ClusterProperties() {
+  const {t} = useTranslation()
   const clusterName = useState(['app', 'clusters', 'selected'])
   const clusterPath = ['clusters', 'index', clusterName]
   const cluster: ClusterDescription = useState(clusterPath)
@@ -90,7 +91,7 @@ export default function ClusterProperties() {
       <Container header={<Header variant="h3">Properties</Header>}>
         <ColumnLayout columns={3} variant="text-grid">
           <SpaceBetween size="l">
-            <ValueWithLabel label="SSH Command">
+            <ValueWithLabel label={t('cluster.properties.sshcommandlabel')}>
               <div className="custom-wrapping">
                 <Box margin={{right: 'xxs'}} display="inline-block">
                   <Popover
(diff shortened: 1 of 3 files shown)


frontend/src/old-pages/Clusters/Properties.tsx
```diff
@@ -100,36 +100,25 @@ export default function ClusterProperties() {
...
-                    cluster.region,
-                  )}/cloudformation/home?region=${
-                    cluster.region
-                  }#/stacks/events?filteringStatus=active&filteringText=&viewNested=true&hideStacks=false&stackId=${
-                    cluster.cloudformationStackArn
-                  }`}
-                >
-                  {cluster.cloudformationStackArn}
-                </Link>
+                ssh {clusterDefaultUser(cluster)}@{headNode.publicIpAddress}
```
mtfranchetto, line 121: question: I'm lacking context on why this feature is needed, especially when there's already an integration with SSM. 
Some questions that can frame the problem better:
- Is this needed to connect to the head node for troubleshooting? 
- Why a user should use SSH over SSM?
- If a user has to connect to the head node, they are going to need a key-pair or some kind of authentication mechanism, so the command, as it is right now, is not usable in a production environment. What's the value of the it?

frontend/src/old-pages/Clusters/Properties.tsx
```diff
@@ -100,36 +100,25 @@ export default function ClusterProperties() {
                     triggerType="custom"
                     content={
                       <StatusIndicator type="success">
-                        ARN copied
+                        SSH Command copied
```
mtfranchetto, line 103: chore: we are moving towards a more robust product and started migrating all hard-coded labels to an i18n framework so the copywriting will be easier to review and translate. That's why I suggest you move all the edited labels to the `strings.json` file and use the `t` function from the `useTranslation` hook to retrieve the values.
An example of how this is done in [this commit](https://github.com/aws-samples/pcluster-manager/commit/a2e2adbc70bedc675457fe4ce6ca74585e10328d).

sean-smith: Thanks I'll do that!

frontend/src/old-pages/Clusters/Properties.tsx
sean-smith: So the idea here is to make it easier for users to use their native SSH client to connect. Yes they will need a keypair, however if they've added their keypair as the default key (either by naming it `id_rsa` or setting that in their ssh config) they can connect using that command. I'd love to include an option to download the keypair also (or upload their pub key in PCM) but that's a much bigger feature.

There's a number of reasons users would want to connect via SSH instead of SSM.
1. They don't have console access. 
2. The cluster isn't setup with SSM access
3. They have a multi-user environment setup and need to authenticate with SSH
4. They want to use their local CLI instead of the browser
5. They want to copy files (using scp or sftp)

frontend/src/old-pages/Clusters/Properties.tsx
sean-smith: done ✅

frontend/src/old-pages/Clusters/Properties.tsx
```diff
@@ -100,36 +101,46 @@ export default function ClusterProperties() {
...
+                <HelpTooltip>
+                  <Trans i18nKey="cluster.properties.sshcommand.tooltiptext">
+                    <a
+                      rel="noreferrer"
+                      target="_blank"
+                      href="https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/AccessingInstancesLinux.html#AccessingInstancesLinuxSSHClient"
+                    ></a>
+                  </Trans>
+                </HelpTooltip>
+                {/* <HelpTooltip>
```
mtfranchetto, line 132: todo: just remove the comment and it's ready to be merged!

sean-smith: done!

PR #394, opened 2022-12-07
Added CSRF implementation for write operation endpoints

BarcoMasile:
## Description
This PR introduces CSRF protection implementation for our write operation endpoints.

Disclaimer: this PR would break Demo env since no CRSF implementation would be on the Frontend, so it needs to wait for the Frontend part.

## Changes
- Refactored before_func that sets the global logger instance to a proper Flask Extension that takes care of right ordering of `before_func`s since it's fundamental for the backend logic and logger especially needs to be available right away
- moved logger instantiation since it shouldn't be exportable and needed in the code thanks to the global LocalProxy
- moved ExceptionHandler
- moved SecurityHeaders
- introduced a centralized module for PCM custom exceptions
- added exception handler for CSRFError
- added fingerprint generator interface + cognito impl + test
- added [`itsdangerous`](https://itsdangerous.palletsprojects.com/en/2.1.x/) dependency (with pre-approved License)
- improved tests folder structure and added centralized fixtures for subfolders `security` and `exceptions`
- added CSRF implementation
- added CSRF unit tests
- adopted CSRF + decorator to protect state changing endpoints
- removed unnecessary comments and import from `PclusterApiHandler.py`
- in `PclusterApiHandler.py`: adopted CSRF decorator + removed unnecessary `authenticate` invocations
- added safeswitch to disable CSRF (in tests) + fixed broken test after csrf implementation

## How Has This Been Tested?
- unit tests
- manual integration tests

## References
- [flask restful method decorators](https://flask-restful.readthedocs.io/en/latest/extending.html#resource-method-decorators#resource-method-decorators)

## PR Quality Checklist


api/PclusterApiHandler.py
```diff
@@ -696,34 +696,30 @@ def _get_params(_request):
 
 
 class PclusterApiHandler(Resource):
-    method_decorators = [authenticated({"admin"})]
+    __read_decorators = [authenticated({'admin'})]
+    __write_decorators = [authenticated({'admin'}), csrf_needed]
+
+    method_decorators = {
+        'get': __read_decorators,
```
mendaomn, line 703: issue: are we sure about this? What's the drawback of checking CSRF on `GET`s as well? 

The fact that `GET`s are non-mutating is not guaranteed unless we trust the way APIs are implemented

app.py
```diff
@@ -80,11 +85,13 @@ def serve(path):
 
     @app.route("/manager/ec2_action", methods=["POST"])
     @authenticated(ADMINS_GROUP)
+    @csrf_needed
```
mendaomn, line 88: suggestion: would it be safer to implement a whitelist instead? For instane, CSRF could be needed by default, and "unprotected" endpoints need to be explicitly decorated

This would also be a future improvement for the `authenticated` decorator

api/security/csrf/csrf.py
```diff
@@ -35,7 +35,7 @@ def is_csrf_enabled():
     return os.getenv("ENABLE_CSRF") != 'false'
 
 def set_csrf_cookie(resp, csrf_token):
-    resp.set_cookie(CSRF_COOKIE_NAME, value=csrf_token)
+    resp.set_cookie(CSRF_COOKIE_NAME, value=csrf_token, httponly=True, secure=True, samesite="Lax")
```
mendaomn, line 38: improvement: it would be important to test the cookie attributes as well (I think we test them in other cookies) since those attributes are important for security

api/exception/handlers.py
```diff
@@ -22,6 +23,10 @@ def value_error_handler(err):
     logger.error(descr, extra=dict(status=code, exception=type(err)))
     return __handler_response(code, descr)
 
+def csrf_error_handler(err):
+    descr, code = str(err), 403
```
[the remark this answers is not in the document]
BarcoMasile: The only CSRFError that get thrown is
`raise CSRFError("provided CSRF are not valid or values do not match")`
so the only info that we put out is the minimum to inform the client of what went wrong
It should be ok

api/PclusterApiHandler.py
BarcoMasile: From my tests it seems that GET calls are actually to read data and not change them (I may have missed something of course)
we could check csrf for GETs as well if we wanted to, I'm open to it

api/security/csrf/csrf.py
BarcoMasile: Done in next revision! Thanks

mtfranchetto (approved):

requirements.txt
```diff
@@ -7,3 +7,4 @@ python-jose==3.3.0
 PyYAML==6.0
 pytest==7.1.2
 pytest-mock==3.8.2
+itsdangerous==2.1.2
```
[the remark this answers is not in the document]
mtfranchetto: Thank you!

app.py
BarcoMasile: I like the approach and we can definitely address this in the near future along the authentication guards

mendaomn (approved):

PR #406, opened 2022-12-12
Configure the Ace library to run in CSP-compliant mode

tmscarla:
## Description

The CloudScape `Code editor` component relies on the Ace library, which receives it as a property. In order to make it CSP-compliant, as specified in the [developer guidelines](https://cloudscape.design/components/code-editor/?tabId=api), we need to set the following configuration parameter to `true`:

```
window.ace.config.set('useStrictCSP', true)
```
This will result in avoiding to include some `<style>` tags into the `<head>` tag, as pointed out in https://github.com/ajaxorg/ace/issues/3260. The two attached images shows the difference of having the property unset / set to `true`.

## How Has This Been Tested?

* Run PCM locally and manually inspected through developer tools

## PR Quality Checklist

image (57)
image (56)

### DIFF
--- a/frontend/src/pages/_app.tsx
+++ b/frontend/src/pages/_app.tsx
@@ -39,6 +39,7 @@ function App({Component, pageProps}: AppProps) {
     window.ace.config.set('loadWorkerFromBlob', false)
     window.ace.config.set('showFoldWidgets', false)
     window.ace.config.set('showPrintMargin', false)
+    window.ace.config.set('useStrictCSP', true)
   }, [])
 
   return (


mendaomn (commented):
LGTM. Aside from following the recommendation of Cloudscape, I couldn't validate this would otherwise break `CodeEditor`. 

Can you just double check there are no negatives of adopting the `useStrictCSP` config? (e.g. performance)

tmscarla:
> LGTM. Aside from following the recommendation of Cloudscape, I couldn't validate this would otherwise break `CodeEditor`.
> 
> Can you just double check there are no negatives of adopting the `useStrictCSP` config? (e.g. performance)

Currently I don't see any drawbacks in setting `useStrictCSP` to true. I have manually tested both configurations and the editor looks fine. We can follow up with our affinity engineer to dive deep into the question.

mendaomn (approved):

PR #409, opened 2022-12-13
Fetch and send CSRF token on every request

mendaomn:
## Description

This PR implements the Double Submit Cookie strategy on the frontend to prevent CSRF attacks.

## Changes

- fetch token via `GET /csrf` on application start
- set the `X-CSRF-Token` header for every subsequent request

## How Has This Been Tested?

- unit tests
- manually

## References

- [Double Submit Cookie](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html#double-submit-cookie)

## PR Quality Checklist

### DIFF
--- a/frontend/src/model.tsx
+++ b/frontend/src/model.tsx
@@ -16,13 +16,15 @@ import {
   setState,
   updateState,
 } from './store'
-import {USER_ROLES_CLAIM} from './auth/constants'
 import {generateRandomId} from './util'
 
 // UI Elements
 import {AppConfig} from './app-config/types'
 import {getAppConfig} from './app-config'
 import {axiosInstance, executeRequest, HTTPMethod} from './http/executeRequest'
+import {getCsrfToken} from './auth/getCsrfToken'
+import {ILogger} from './logger/ILogger'
+import {setCsrfTokenHeader} from './http/setCsrfTokenHeader'
 
 // Types
 type Callback = (arg?: any) => void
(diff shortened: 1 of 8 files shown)


frontend/src/model.tsx
```diff
@@ -956,10 +958,23 @@ async function GetAppConfig() {
...
+    logger.error('Could not fetch CSRF token')
+    throw error
+  }
+}
+
+async function LoadInitialState(logger: ILogger) {
   const region = getState(['app', 'selectedRegion'])
   clearState(['app', 'aws'])
   clearAllState()
+  await LoadCsrfToken(logger)
```
mtfranchetto, line 977: question: failing this call will prevent everything that comes later to load up. Have you considered using at least a `notify` call to inform the user something bad has happened?

mtfranchetto (approved):

frontend/src/model.tsx
mendaomn: An error in fetching the CSRF token may be hard to handle (not much we can do aside from retrying indefinitely)

However, subsequent calls would fail with a specific error code. I'll add an handler to `notify` the user and log the error when the CSRF check fails

mendaomn: Coming back to this. The backend is going to return 403 Forbidden (which is already handled), I'll add an error message to the `LoadCsrfToken` command

mtfranchetto (approved):

PR #417, opened 2022-12-16
Added cloudscape breadcrumbs for top level pages

BarcoMasile:
## Description
Added cloudscape breadcrumbs for top level pages

## Changes
- added breadcrumbs prop to AppLayout
- added helper functions to build proper Breadcrum Items
- added hardcoded breadcrumb items to each top level page + cluster create/update info
- added margin top to cluster create/edit Box component to fix layout issue

## How Has This Been Tested?
- manually

## References
- [Breadcrumb Group](https://cloudscape.design/components/breadcrumb-group/?tabId=playground)


frontend/src/old-pages/Configure/Configure.tsx
```diff
@@ -301,22 +308,26 @@ function Configure() {
...
-            {text: 'Clusters', href: '#clusters'},
-            // @ts-expect-error TS(2741) FIXME: Property 'href' is missing in type '{ text: string... Remove this comment to see the full error message
-            {text: editing ? 'Update' : 'Create'},
-          ]}
+    <Layout
+      contentType="form"
+      breadcrumbs={
+        <Breadcrumbs
           onClick={clearStateAndCloseWizard}
+          pages={[
```
mtfranchetto, line 316: todo: memoize this array with `useMemo` since you are using a variable here. For all the other places a simple extract from component + `const` is sufficient

frontend/src/old-pages/Configure/Configure.tsx
```diff
@@ -301,22 +308,26 @@ function Configure() {
...
+            breadcrumbItemFromSlug('clusters'),
+            breadcrumbItem(editing ? 'Update' : 'Create'),
+          ]}
         />
+      }
+    >
+      {/*<Container >*/}
```
mtfranchetto, line 323: nitpick: unneeded comment

frontend/src/old-pages/Configure/Configure.tsx
```diff
@@ -301,22 +308,26 @@ function Configure() {
...
+      <StopDialog clusterName={clusterName} />
+      <SpaceBetween direction="vertical" size="l">
```
mtfranchetto, line 325: question: why are you using two nested `SpaceBetween`? I'm struggling to understand the behavior here

frontend/src/old-pages/Configure/Configure.tsx
```diff
@@ -301,22 +308,26 @@ function Configure() {
...
         <SpaceBetween direction="horizontal" size="s">
           <SideNav />
           <div style={{minWidth: '800px', maxWidth: '1000px'}}>
             <SpaceBetween direction="vertical" size="s">
-              <Box>
+              <Box margin={{top: 'l'}}>
```
mtfranchetto, line 330: todo: avoid using inline style (for the same reason as `useMemo`): the styles can be extracted outside the component with a simple `const`

frontend/src/old-pages/Layout.tsx
```diff
@@ -38,6 +39,7 @@ export default function Layout({
         content={children}
         contentType="table"
         navigation={<SideBar />}
+        breadcrumbs={breadcrumbs}
```
mendaomn, line 42: suggestion: you can centralize implementation of Breadcrumb by putting it here, and just receiving the current breadcrumb "slug"

Then you use your map to:
- find the key for the name of the page, and run it through `i18n`
- find the href (as you are already doing)

frontend/src/old-pages/Configure/Configure.tsx
```diff
@@ -301,22 +308,26 @@ function Configure() {
...
-          ]}
+    <Layout
+      contentType="form"
+      breadcrumbs={
+        <Breadcrumbs
           onClick={clearStateAndCloseWizard}
+          pages={[
+            breadcrumbItemFromSlug('clusters'),
+            breadcrumbItem(editing ? 'Update' : 'Create'),
+          ]}
```
mendaomn, line 319: issue: unless you are memoizing everything properly, creating a new array at every render could cause excessive re-renders

As stated in the other comments, one solution could be:
- just pass a key to the Layout component (e.g. `clusters-update`)
- the Layout component could use `useMemo` if needed (double check if re-renders cause a rerender of the full layout)

frontend/src/pages/breadcrumbs.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export function breadcrumbItemFromSlug(slug: Page): BreadcrumbGroupProps.Item {
+  return pageBreadcrumbItems[slug]
+}
+
+export function breadcrumbItem(text: string): BreadcrumbGroupProps.Item {
+  return {text, href: '#'}
+}
+
+const mainBreadcrumItem = {text: 'ParallelCluster Manager', href: '/clusters'}
```
mtfranchetto, line 22: nitpick: I think you can route to `/`. There's a redirection to `/clusters` inside the router so, again, we don't replicate the routing behavior here.

frontend/src/pages/breadcrumbs.tsx
```diff
@@ -0,0 +1,40 @@
+import BreadcrumbGroup from '@cloudscape-design/components/breadcrumb-group'
+import {BreadcrumbGroupProps} from '@cloudscape-design/components/breadcrumb-group/interfaces'
+import {CancelableEventHandler} from '@cloudscape-design/components/internal/events'
+
+type Page = 'clusters' | 'custom-images' | 'official-images' | 'users'
+
+const pageBreadcrumbItems: Record<Page, BreadcrumbGroupProps.Item> = {
+  clusters: {text: 'Clusters', href: '#clusters'},
```
mendaomn, line 8: issue: as suggested in the other comment, please add i18n to the mix

frontend/src/pages/breadcrumbs.tsx
```diff
@@ -0,0 +1,40 @@
+import BreadcrumbGroup from '@cloudscape-design/components/breadcrumb-group'
+import {BreadcrumbGroupProps} from '@cloudscape-design/components/breadcrumb-group/interfaces'
+import {CancelableEventHandler} from '@cloudscape-design/components/internal/events'
+
+type Page = 'clusters' | 'custom-images' | 'official-images' | 'users'
+
+const pageBreadcrumbItems: Record<Page, BreadcrumbGroupProps.Item> = {
```
mtfranchetto, line 7: suggestion(non-blocking): replicating the same routing structure here for breadcrumbs it's not the best thing to do IMHO,  and i t has to be in sync with changes on existing and new routes.
What you can do here is to take the `path` from `useLocation`, and use it to build a default breadcrumb inside the `Layout` component, which can be overridden with a prop (for example in the `Clusters` case).
If you follow this approach there's no need to enumerate the routes here, and you get a breadcrumb automatically as soon as a new page is added via the `Layout` component.
Example:
```
const location = useLocation()
const breadCrumbText = _.capitalize(location.pathname.split('/')?.[1].replace('-', ' '))
```

mtfranchetto (changes requested):

frontend/src/pages/breadcrumbs.tsx
mendaomn: I like the configuration approach a little bit better because:
- we should make sure we don't just reflect navigations, but the information architecture ([see here](https://cloudscape.design/components/breadcrumb-group/?tabId=usage)). Not a big deal in our case
- we should still i18n-ize the strings (which we can do with your approach as well, just less idal)

My suggestion (see other comments) would be to just associate every page with a key and let the config do the rest, but I'll leave it up to you guys as I don't see strong arguments for either approach

mtfranchetto: +1 for the information architecture, thank you! Resolving the comment

frontend/src/pages/breadcrumbs.tsx
BarcoMasile: Thanks, done in next revision!

frontend/src/old-pages/Configure/Configure.tsx
BarcoMasile: I don't know why GH shows it as added by this PR, but I actually didn't add any `SpaceBetween` components
This is as it was before so I wouldn't be able to answer :-/

frontend/src/old-pages/Configure/Configure.tsx
BarcoMasile: Oopsie, done in next revision!

frontend/src/pages/breadcrumbs.tsx
BarcoMasile: added in the next rev (coming up)

frontend/src/old-pages/Layout.tsx
BarcoMasile: Did as you said, thanks for the suggestion! Next rev is coming up!

frontend/src/old-pages/Configure/Configure.tsx
BarcoMasile: Done (and I also centralized the implementation so this snippet got simpler), I used `useMemo` to avoid re-renders.
Thanks!

frontend/src/old-pages/Layout.tsx
```diff
@@ -43,6 +114,7 @@ export default function Layout({
         content={children}
         contentType="table"
         navigation={<SideBar />}
+        breadcrumbs={<Breadcrumbs slug={pageSlug} {...slugOnClick} />}
```
mendaomn, line 117: issue: I think this breaks the `onClick`, you should be able to see it for yourself by just typing the `slugOnClick` Prop differently than with `any`

In short, you are spreading a function, you should set `onClick={slugOnClick}`

BarcoMasile: Fixed

mendaomn (approved):

mtfranchetto (approved):

PR #422, opened 2022-12-20
Version

sean-smith:
## Description

image

Display ParallelCluster API version in the top bar. This helps customers identify the version that PCM is deployed with and allows them to easily see which clusters they can update.

## Changes

* Display ParallelCluster API version in the top bar

## Changelog entry

* Display ParallelCluster API version in the top bar

## How Has This Been Tested?

Tested by deploying manually

## PR Quality Checklist

### DIFF
--- a/frontend/locales/en/strings.json
+++ b/frontend/locales/en/strings.json
@@ -1,5 +1,10 @@
 {
   "global": {
+    "topbar": {
+      "version": {
+        "ariaLabel": "Version of AWS ParallelCluster API"
+      }
+    },
     "menu": {
       "header": "ParallelCluster Manager",
       "clusters": "Clusters",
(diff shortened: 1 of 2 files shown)


mtfranchetto:
Hi @sean-smith,
while I understand the problem you're trying to solve with this this pull request, I don't think the solution implemented here fits very well in PCM and its future evolution because:
- it's a piece of info that's not strictly required to be always visible, and it could be restricted to cluster creation or edit
- placed in the `TopBar` it becomes a button, which shouldn't be in this case as it's a mere text info, with wrong accessibility hints

For these reasons I'm going to close this pull request. Feel free to engage with both the product and design team to discuss a proper solution for this problem.
Thank you

PR #426, opened 2022-12-21
Add stacktraces to Lambda logs

mtfranchetto:
## Description

Print the complete stacktrace when logging an error on CW.

Although it's not ideal to have each stacktrace entry on a separate line (and I thought this was an issue) it's actually the behavior of the Python Lambda runtime, and cannot be changed (more details [here](https://docs.aws.amazon.com/lambda/latest/dg/python-logging.html)).
This can behavior can be overcome by going through Logs Insights first to find the log stream and time of the error, then visiting the full log stream to see the stacktrace.

## How Has This Been Tested?

Screenshot 2022-12-21 at 12 46 33

## PR Quality Checklist

- [] I removed hardcoded strings and used our `i18n` solution instead (see [here](https://github.com/aws-samples/pcluster-manager/pull/175/commits/fdc6b77987c87a26f51dbc8da5d371d95ef80601))

### DIFF
--- a/api/logger.py
+++ b/api/logger.py
@@ -27,9 +27,7 @@ class DefaultLogger(object):
     self.logger.warning(self._log_output(msg, extra))   
     
   def error(self, msg, extra=None):
-    self.logger.error(self._log_output(msg, extra))
+    self.logger.error(self._log_output(msg, extra), exc_info=True)
 
   def critical(self, msg, extra=None):
-    self.logger.critical(self._log_output(msg, extra))
-
-
+    self.logger.critical(self._log_output(msg, extra), exc_info=True)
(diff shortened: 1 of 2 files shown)


api/logger.py
```diff
@@ -27,9 +27,7 @@ def warning(self, msg, extra=None):
     self.logger.warning(self._log_output(msg, extra))   
     
   def error(self, msg, extra=None):
-    self.logger.error(self._log_output(msg, extra))
+    self.logger.error(self._log_output(msg, extra), exc_info=True)
```
BarcoMasile, line 30: nitpick(non-blocking, super small): since this would log 'NoneType: None' every type an exception was not previously raised, this could be edited to print Exception info only if an exception was actually there.
```suggestion
    self.logger.error(self._log_output(msg, extra), exc_info=self.__is_exception_caught())
    
    def __is_exception_caught(self):
      return sys.exc_info() != (None, None, None)
```

BarcoMasile (dismissed):
Approved, the only thing I would keep in mind is that, if running still in an `except` clause, we could use `logger.error` again and print info about a previously caught exception even if that's not the intention

mtfranchetto:
> Approved, the only thing I would keep in mind is that, if running still in an `except` clause, we could use `logger.error` again and print info about a previously caught exception even if that's not the intention

Makes sense, but this case is not that relevant to change the implementation right now. By the way, let's keep an eye and, if that happens, let's figure out how to change the logger.

BarcoMasile (approved):

PR #430, opened 2022-12-22
Bubble up boto3 Cognito exception for delete user operation

BarcoMasile:
## Description
Removed the try-except clause from the delete operation in order to bubble up the real Cognito exception raised so our Boto exception handler can handle it the right way.

## How Has This Been Tested?
- manually
- existing unit tests

## PR Quality Checklist

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -600,14 +600,10 @@ def list_users():
 
 
 def delete_user():
-    try:
-        cognito = boto3.client("cognito-idp")
-        username = request.args.get("username")
-        cognito.admin_delete_user(UserPoolId=USER_POOL_ID, Username=username)
-        return {"Username": username}
-    except Exception as e:
-        return {"message": str(e)}, 500
-
+    cognito = boto3.client("cognito-idp")
+    username = request.args.get("username")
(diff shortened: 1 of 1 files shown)


api/PclusterApiHandler.py
```diff
@@ -600,14 +600,10 @@ def list_users():
...
-        cognito = boto3.client("cognito-idp")
-        username = request.args.get("username")
-        cognito.admin_delete_user(UserPoolId=USER_POOL_ID, Username=username)
-        return {"Username": username}
-    except Exception as e:
-        return {"message": str(e)}, 500
-
+    cognito = boto3.client("cognito-idp")
+    username = request.args.get("username")
+    cognito.admin_delete_user(UserPoolId=USER_POOL_ID, Username=username)
```
lazyoft, line 605: question: if we remove the exception handler wouldn't this call be intercepted by the global exception handler? I might be OK with this, but in that case I think that the title of both the PR and the commit are misleading, as in this case it should be something like "bubble up cognito exceptions"

BarcoMasile: The botocore handler works differently than the others because it gets the error code from what boto returned (which is usually correct), and works for every other exception that boto may raise.
I'll change the PR title right away

lazyoft (approved):

PR #441, opened 2022-12-27
Handle buffered log entries on the backend

BarcoMasile:
## Description
Changed `/logs` controller implementation to allow buffered logs processing instead of one entry at a time.
Moved `source` info handling to the frontend code.

## Changes
Right now the controller accepts a `"application/json"` content of a list of log entries.
Tests have been adjusted, since the implementation is pretty simple I edited the existing tests without separating controller and service tests since it wouldn't add value.
Moved logging related code to a new package `logging`.
Removed `source` extra key handling from the backend code and moved to the frontend code, refactoring the ILogger interface and implementation to allow the passing of the new parameter (as optional).

## How Has This Been Tested?
- manually
- unit tests run on my local machine
- e2e tests run on my local machine

## PR Quality Checklist


app.py
```diff
@@ -184,24 +184,14 @@ def logout_():
     @authenticated(ADMINS_GROUP)
     @csrf_needed
     def push_log():
-        if 'level' not in request.json or 'message' not in request.json:
-            raise ValueError('Request body missing one or more mandatory fields ["message", "level"]')
+        if type(request.json) != list:
```
lazyoft, line 187: question: why are we not accepting a single entry anymore?

api/logging/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+    if not (extra is None or type(extra) is dict):
+        raise ValueError('Extra param must be a valid json object')
+
+    lowercase_level = level.lower()
+    if lowercase_level not in VALID_LOG_LEVELS:
+        raise ValueError('Level param must be a valid log level')
+
+    return lowercase_level, message, extra
+
+def log_from_frontend(_logger, level, message, extra):
```
lazyoft, line 27: question: why do we need to be as specific as implying a log is coming from the frontend in this case?

BarcoMasile: the reason is that this function takes care of logging implementing a custom logic that is only valid for logs coming from the frontend (which right now is only doing `extra['source'] = 'frontend'` but may change in the future).
In the BE code we call the logger directly

app.py
BarcoMasile: Switching to a buffering logger all log entries would go through the buffering process so the previous logic would not be needed anymore. Of course we can easily reintroduce it if we need it again

api/logging/__init__.py
lazyoft: I will leave the considerations on the naming and purpose to other reviewers.

BarcoMasile: After some thought I renamed the function to `push_log_entry` and it now accepts a new parameter (mandatory) called `source` to let the caller specify the source of the log, even though right now that value is always "frontend"

app.py
```diff
@@ -183,24 +184,14 @@ def logout_():
...
+            raise ValueError('Body must be a list of log entries')
```
mtfranchetto, line 188: nitpick: by looking at Python built-in exceptions this should be a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError).

> Passing arguments of the wrong type (e.g. passing a [list](https://docs.python.org/3/library/stdtypes.html#list) when an [int](https://docs.python.org/3/library/functions.html#int) is expected) should result in a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError), but passing arguments with the wrong value (e.g. a number outside expected boundaries) should result in a [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

api/logging/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+    if lowercase_level not in VALID_LOG_LEVELS:
+        raise ValueError('Level param must be a valid log level')
+
+    return lowercase_level, message, extra
+
+def push_log_entry(_logger, level, message, extra, source):
+    """ Logs a single log entry coming from the fronted """
+    logging_fun = getattr(_logger, level, None)
+    extra = {} if extra is None else extra
+    extra['source'] = source
```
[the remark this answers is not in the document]
BarcoMasile: It can, could be done in another PR

app.py
BarcoMasile: Yes, but that would be appropriate for a library function exception
I chose the ValueError because the "value" received by the API is not correct
Basically I'm using ValueErrors as API validation error

mtfranchetto (approved):

PR #454, opened 2023-01-03
Show opt-in region in topbar dropdown only if PCM is deployed in it

tmscarla:
## Description
Currently PCM displays the full list of [regions supported by PC](https://docs.aws.amazon.com/parallelcluster/latest/ug/supported-regions-v3.html) in its topbar dropdown menu. If an opt-in region is selected, and the account where PCM is deployed has not opted-in yet for the selected region, a series of errors is thrown on the frontend.

This PR aims to prevent these errors from being thrown by showing opt-in regions in topbar dropdown only if PCM is deployed in it. In the attached example PCM has been deployed in an opt-in region (Milan, eu-south-1) which is shown at the top of the list.

## Changelog entry

* Prevent frontend errors by showing opt-in regions in topbar dropdown only if PCM is deployed in it

## How Has This Been Tested?
1. Opted-in `eu-south-1` region in my personal account where PCM is deployed
2. Verified that `eu-south-1` region is displayed at the top of the regions list

## References

* [PC supported regions](https://docs.aws.amazon.com/parallelcluster/latest/ug/supported-regions-v3.html)
* [AWS Opt-in regions](https://docs.aws.amazon.com/general/latest/gr/rande-manage.html)

Screenshot 2023-01-03 at 12 32 32

## PR Quality Checklist


frontend/src/components/TopBar.tsx
```diff
@@ -43,25 +40,27 @@ function regions(selected: any) {
...
+    ['Asia Pacific (Hong Kong)', 'ap-east-1'],
+    ['Europe (Milan)', 'eu-south-1'],
+    ['Middle East (Bahrain)', 'me-south-1'],
+    ['China (Beijing)', 'cn-north-1'],
+    ['China (Ningxia)', 'cn-northwest-1'],
+    ['AWS GovCloud (US-East)', 'us-gov-east-1'],
+    ['AWS GovCloud (US-West)', 'us-gov-west-1'],
+  ]
+
+  const optInRegion = optInRegions.filter(region => region[1] === selected)
```
lazyoft, line 60: question: this seems to show only the opt-in region in which PCM has been installed. What if I have it installed in another region but I want to create clusters in an opt-in one? Shouldn't we support all the regions a customer has opted-in rather than just the one he might have installed PCM?

frontend/src/components/TopBar.tsx
```diff
@@ -21,18 +21,15 @@ import {useLogger} from '../logger/LoggerProvider'
 import {NavigateFunction, useNavigate} from 'react-router-dom'
 
 function regions(selected: any) {
```
mendaomn, line 23: suggestion: Let's add a test to this function, since we are adding some business logic

frontend/src/components/TopBar.tsx
mendaomn: You are correct, support for other opt-in regions is out of the scope of this PR (more details in the related task tracker)

frontend/src/components/TopBar.tsx
tmscarla: Will add in next revision :)

mendaomn (approved):

PR #456, opened 2023-01-03
Added support for configuring existing PCM-deployed Cognito user pool with a new PCM installation

BarcoMasile:
## Description
This PR adds support for setting up a new deployment of PCM without deploying a new Cognito user pool instance, so that a previously deployed user pool could be used instead.
This is achieved through the use of new parameters (easily obtainable from CloudFormation console) to be passed at stack creation time.

## Changes
- Introduced external cognito parameters
- adjusted parameter descriptions

## How Has This Been Tested?
- manual tests on multiple deployments on my personal account

## PR Quality Checklist

### DIFF
--- a/infrastructure/pcluster-manager.yaml
+++ b/infrastructure/pcluster-manager.yaml
@@ -1,14 +1,31 @@
 Parameters:
   AdminUserEmail:
-    Description: Email address of administrative user setup by default.
+    Description: Email address of administrative user to setup by default (only with new Cognito instances).
     Type: String
-    MinLength: 1 # Force this to be selected
   PublicEcrImageUri:
     Description: When specified, the URI of the Docker image for the Lambda of the PclusterManager container
     Type: String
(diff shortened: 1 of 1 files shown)


infrastructure/pcluster-manager.yaml
```diff
@@ -1,14 +1,31 @@
 Parameters:
   AdminUserEmail:
-    Description: Email address of administrative user setup by default.
+    Description: Email address of administrative user to setup by default (only with new Cognito instances).
     Type: String
-    MinLength: 1 # Force this to be selected
```
mtfranchetto, line 5: question: is it possible to validate this constraint only when creating a new Cognito stack?

mtfranchetto (approved):

infrastructure/pcluster-manager.yaml
BarcoMasile: Unfortunately I couldn't find anything like that, so I added the info in the description `(only with new Cognito instances)` to this field

PR #464, opened 2023-01-09
Added OAuth2 refresh token flow implementation + shortened `id` and `access` token lifespans

BarcoMasile:
## Description
This PR introduces the implementation of the refresh token flow for the OAuth2 protocol, enabling us to shorten token lifespans without breaking user experience.

## Changes
- the `authenticate` method now checks if it's possible to refresh tokens when an expired accesstoken is used, and if so it performs the refresh setting new cookies values to be added to the response in a request context ( `g.auth_cookies`) to be used in a custom `afterresponse` func (this was done to avoid handling response objects manually since a view function could return either a dict and a code like in `app.py`, or a true `Response` object, like in `handlers.py`)
- added custom refreshtokenerror exception + handler
- tests have been added and existing tests adjusted
- edited the pcluster-manager.yaml to shorten lifespans of id and access token

## How Has This Been Tested?
- unit tests
- manual tests
- manually deployed PCM in cloudformation to verify shortened ilfespans

## PR Quality Checklist

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -22,6 +22,8 @@ from flask import abort, redirect, request
 from flask_restful import Resource, reqparse
 from jose import jwt
 
+from api.exception.exceptions import RefreshTokenError
+from api.pcm_globals import set_auth_cookies_in_context
 from api.security.csrf.constants import CSRF_COOKIE_NAME
 from api.security.csrf.csrf import csrf_needed
 from api.utils import disable_auth
@@ -116,8 +118,24 @@ def sigv4_request(method, host, path, params={}, headers={}, body=None):
 
     return req_call(boto_request.url, data=body_data, headers=boto_request.headers, timeout=30)
 
+def refresh_tokens(refresh_token):
+    auth = requests.auth.HTTPBasicAuth(CLIENT_ID, CLIENT_SECRET)
 
-# Wrappers
+    resp = requests.post(
+        TOKEN_URL,
+        data={"grant_type": 'refresh_token', "refresh_token": refresh_token, "client_id": CLIENT_ID},
+        auth=auth,
+        headers={"Content-Type": "application/x-www-form-urlencoded"},
+    )
+
+    if resp.status_code != 200:
+        raise RefreshTokenError(resp.json().get('error'))
+
+    values = resp.json()
+    access_token = values.get("access_token")
+    id_token = values.get("id_token")
+
+    return {'accessToken': access_token, 'idToken': id_token}
 
 def authenticate(groups):
     if disable_auth():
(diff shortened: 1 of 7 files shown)


api/PclusterApiHandler.py
```diff
@@ -116,8 +118,24 @@ def sigv4_request(method, host, path, params={}, headers={}, body=None):
...
 
-# Wrappers
+    if resp.status_code != 200:
+        raise RefreshTokenError(resp.json().get('error'))
+
+    values = resp.json()
+    access_token = values.get("access_token")
+    id_token = values.get("id_token")
+
+    return {'accessToken': access_token, 'idToken': id_token}
```
mendaomn, line 138: question: why do we not expect a new refresh token?(I'm reading [this for reference](https://auth0.com/docs/secure/tokens/refresh-tokens/refresh-token-rotation), and it seems the flow involves returning a new refresh token)

api/PclusterApiHandler.py
```diff
@@ -130,7 +148,13 @@ def authenticate(groups):
     try:
         decoded = jwt_decode(access_token)
     except jwt.ExpiredSignatureError:
-        return abort(401)
+        refresh_token = request.cookies.get('refreshToken', None)
+        if refresh_token is None:
+            return abort(401)
```
mendaomn, line 153: suggestion: I believe this case is already covered by the `refresh_tokens` function (I'm guessing it will raise a `RefreshTokenError`)

infrastructure/pcluster-manager.yaml
```diff
@@ -259,11 +259,11 @@ Resources:
...
       RefreshTokenValidity: 7
-      AccessTokenValidity: 1
-      IdTokenValidity: 1
+      AccessTokenValidity: 10
+      IdTokenValidity: 10
       TokenValidityUnits:
-        AccessToken: "days"
-        IdToken: "days"
+        AccessToken: "minutes"
+        IdToken: "minutes"
```
mendaomn, line 267: question: have you checked `update_infrastructure.sh` can successfully run after these changes?

api/tests/test_authenticate.py
```diff
@@ -59,6 +63,18 @@ def test_authenticate_with_signature_exception_returns_401(mocker, app):
 
         mock_abort.assert_called_once_with(401)
 
+def test_authenticate_with_valid_refresh_token(mocker, app):
```
mendaomn, line 66: nit-pick: I believe it would be slightly more informative to state it like "test authenticate with expired token" or something along those lines (we are not testing the validity of the refresh token)

BarcoMasile: Nice, I'll follow your advice

infrastructure/pcluster-manager.yaml
BarcoMasile: I haven't run the command so unfortunately no (but I will of course check for the outcome and fix in case)

api/PclusterApiHandler.py
BarcoMasile: Cognito doesn't return a new refresh token (as other OIDC providers do) but instead you need to keep the old one and reuse it until it expires

api/PclusterApiHandler.py
BarcoMasile: As it is right now, refresh_tokens performs the POST and expects a refresh token
I preferred to leave the concern of checking for the presence of the token outside the function that uses it

mendaomn (approved):

PR #465, opened 2023-01-09
Adopt CloudScape Wizard component

tmscarla:
## Description
This PR introduces the CloudScape [Wizard](https://cloudscape.design/components/wizard/?tabId=playground) component in the create cluster wizard.

## Changelog entry

* Adopted CloudScape Wizard component in the create cluster wizard

## How Has This Been Tested?
* Created new cluster
   * Navigated through wizard (next, back, select step)
   * Enable/disable MultiUser and verified that navigation works as expected
   * Dry-run successful
   
* Updated existing cluster
   * Verified options are loaded and validated correctly 
   * Dry-run successful

## References

* https://cloudscape.design/components/wizard/

## PR Quality Checklist

Screenshot 2023-01-09 at 09 53 44

### DIFF
--- a/frontend/src/old-pages/Configure/Configure.tsx
+++ b/frontend/src/old-pages/Configure/Configure.tsx
@@ -8,30 +8,19 @@
 // or in the "LICENSE.txt" file accompanying this file. This file is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES
 // OR CONDITIONS OF ANY KIND, express or implied. See the License for the specific language governing permissions and
 // limitations under the License.
-import React, {useCallback, useMemo} from 'react'
+import React, {useCallback} from 'react'
 import {useNavigate} from 'react-router-dom'
 
-// @ts-expect-error TS(7016) FIXME: Could not find a declaration file for module 'js-y... Remove this comment to see the full error message
-import jsyaml from 'js-yaml'
-
-import {
-  clearState,
-  getState,
-  setState,
-  useState,
-  updateState,
-} from '../../store'
+import {clearState, getState, setState, useState} from '../../store'
 import {LoadAwsConfig} from '../../model'
 
-// UI Elements
 import {
-  Box,
   Button,
-  SideNavigation,
   SpaceBetween,
+  Wizard,
+  WizardProps,
 } from '@cloudscape-design/components'
 
-// SubPages
 import {Source, sourceValidate} from './Source'
 import {Cluster, clusterValidate} from './Cluster'
 import {HeadNode, headNodeValidate} from './HeadNode'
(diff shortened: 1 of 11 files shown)


frontend/src/old-pages/Configure/Configure.tsx
```diff
@@ -302,78 +125,145 @@ function Configure() {
...
+          </Button>
+        )}
+      </SpaceBetween>
+    )
+  }, [editing, fleetStatus, handleRefresh])
+
+  const isLoadingNextStep = useCallback(() => {
+    const aws = getState(['aws'])
+    return !aws
+  }, [getState])
```
mendaomn, line 184: issue: is this supposed to be updated when the state changes? If so, use `useState` instead of `getState`

Also FYI, there is no need to use `useCallback` here (nor `useMemo`) since you are only interested in the return value, and it's a boolean value (therefore no need to memoize at all)

frontend/src/old-pages/Configure/Configure.tsx
```diff
@@ -302,78 +125,145 @@ function Configure() {
...
+
+  const handleOnNavigate: NonCancelableEventHandler<WizardProps.NavigateDetail> =
+    useCallback(
+      ({detail}) => {
+        navigateWizard(detail.reason, detail.requestedStepIndex)
+      },
+      [navigateWizard],
+    )
+
+  const showSecondaryActions = useCallback(() => {
```
mendaomn, line 149: nit-pick: no need to adopt `useCallback` here, since you are interested in the returned value

I think we can get away without memoizing this component for now

frontend/src/old-pages/Configure/Configure.tsx
tmscarla: Going to remove entirely the method in favour of `refreshing` value.

frontend/src/old-pages/Configure/Configure.tsx
```diff
@@ -302,78 +125,145 @@ function Configure() {
...
+
+  const isLoadingNextStep = useCallback(() => {
+    const aws = getState(['aws'])
+    return !aws
+  }, [getState])
+
+  const activeStepindex = useMemo(
+    () => pages.indexOf(currentPage),
+    [currentPage],
+  )
```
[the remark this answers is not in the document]
tmscarla: Gotcha, going to use `pages.indexOf(currentPage)` in the Wizard props directly.

frontend/src/old-pages/Configure/Configure.tsx
tmscarla: Ok, will remove the memoization for now

PR #467, opened 2023-01-09
Added refresh token revocation towards Cognito when logging out

BarcoMasile:
## Description
Added refresh token revocation towards Cognito when logging out

## Changes
- logout function now performs a post request to `${cognitoAuthPath}/oauth2/revoke` to revoke the current valid refresh token so it cannot be used anymore, log a warning in case of failure since it should not disrupt UX

## PR Quality Checklist

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -23,7 +23,7 @@ from flask_restful import Resource, reqparse
 from jose import jwt
 
 from api.exception.exceptions import RefreshTokenError
-from api.pcm_globals import set_auth_cookies_in_context
+from api.pcm_globals import set_auth_cookies_in_context, logger
 from api.security.csrf.constants import CSRF_COOKIE_NAME
 from api.security.csrf.csrf import csrf_needed
 from api.utils import disable_auth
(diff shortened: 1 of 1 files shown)


api/PclusterApiHandler.py
```diff
@@ -673,13 +674,30 @@ def login():
...
+        revoke_cognito_refresh_token(refresh_token)
+
     resp = __cognito_logout_redirect(get_app_config())
     resp.set_cookie("accessToken", "", expires=0)
     resp.set_cookie("idToken", "", expires=0)
     resp.set_cookie("refreshToken", "", expires=0)
     resp.set_cookie(CSRF_COOKIE_NAME, "", expires=0)
     return resp
 
+def revoke_cognito_refresh_token(refresh_token):
```
mendaomn, line 688: issue: let's adapt the logout test to check this call is being performed

api/PclusterApiHandler.py
```diff
@@ -673,13 +674,30 @@ def login():
 
 
 def logout():
+    refresh_token = request.cookies.get('refreshToken', None)
+    if refresh_token is not None:
+        revoke_cognito_refresh_token(refresh_token)
```
mendaomn, line 679: question: is this blocking the response? meaning, will the client (FE) need to wait on the new POST request, before getting a response?

mendaomn:
It's not necessary for this PR, but it would be a good time to add a simple e2e test to check logout works properly, I don't think we have one as of yet

api/PclusterApiHandler.py
BarcoMasile: Yes unfortunately
I didn't make it async because running in a lambda, we would still need to wait for the call to complete or we may interrupt the request when lambda "turns down"

api/PclusterApiHandler.py
BarcoMasile: Added in this new revision, also added tests for the function itself

PR #468, opened 2023-01-09
Limit invocations on API Gateway routes

mtfranchetto:
## Description

By default API Gateway has large throttling limits: this change ensures a reasonable limit is used instead.

## How Has This Been Tested?

- created a new env and verified a user don't get throttled with those limits

## PR Quality Checklist

### DIFF
--- a/infrastructure/pcluster-manager.yaml
+++ b/infrastructure/pcluster-manager.yaml
@@ -237,6 +237,9 @@ Resources:
       ApiId: !Ref ApiGateway
       StageName: $default
       AutoDeploy: True
+      DefaultRouteSettings:
+        ThrottlingBurstLimit: 50 
+        ThrottlingRateLimit: 100
 
   CognitoAppClient:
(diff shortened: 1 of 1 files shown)


infrastructure/pcluster-manager.yaml
```diff
@@ -237,6 +237,9 @@ Resources:
       ApiId: !Ref ApiGateway
       StageName: $default
       AutoDeploy: True
+      DefaultRouteSettings:
+        ThrottlingBurstLimit: 50 
+        ThrottlingRateLimit: 100
```
lazyoft, line 242: question: are these limits suitable for our use case? Should they be higher?

mtfranchetto: Since the product is installed on the customer account and used mostly by cluster admins (and not end users) these limits should be good!

BarcoMasile (approved):

PR #470, opened 2023-01-09
Move Multi-user under Cluster section of the wizard

tmscarla:
## Description

In order to enhance PCUI user experience, we need to thin out the steps of our cluster creation wizard. Following UX design, the whole Multi-user step needs to be removed, and Multi-user section needs to be placed under `Cluster` step of the wizard.

This PR aims at solving the issue, by conditionally showing Multi-user properties in Cluster step through a dedicated checkbox.

## Changelog entry

* Remove `Multi-user` wizard step and embedded it into `Cluster` step in a dedicated section

## How Has This Been Tested?
* Unit tests
* E2E tests
* Manually navigated through the end of the wizard
   * With multi-user enabled
   * With multi-user disabled

## References
* https://cloudscape.design/components/checkbox/

## PR Quality Checklist

Screenshot 2023-01-09 at 17 45 40


frontend/src/old-pages/Configure/__tests__/useWizardNavigation.test.tsx
```diff
@@ -157,122 +157,6 @@ describe('Given a hook to navigate through the wizard', () => {
     })
   })
 
-  describe('when on cluster page', () => {
```
BarcoMasile, line 160: question: the tests deleted from here, shouldn't they be still present somewhere else?

frontend/src/old-pages/Configure/Cluster.tsx
```diff
@@ -404,40 +415,44 @@ function Cluster() {
...
-    </Container>
+    <SpaceBetween direction="vertical" size="s">
+      <Container>
+        <SpaceBetween direction="vertical" size="s">
+          <RegionSelect />
+          <OsSelect />
+          <VpcSelect />
+          <CustomAMISettings
+            basePath={configPath}
+            appPath={['app', 'wizard']}
```
BarcoMasile, line 436: suggestion(non-blocking): this array `['app', 'wizard']` can easily be put outside the re-rendered template. This way you would pass the same reference avoiding useless re-renders

mtfranchetto (approved):

frontend/src/old-pages/Configure/__tests__/useWizardNavigation.test.tsx
tmscarla: It covered the `multiUser` case: if enabled, you navigate to multiUser page, headNode otherwise.

Moving multiUser to cluster page eliminates this edge case.
Standard flow has already been tested in other test cases.

PR #472, opened 2023-01-10
Added API signature validation

BarcoMasile:
## Description
Added API input validation mechanism + decorator to validate API invocation at the controller level.

## Changes
- added Marshmallow dependency to `requirements.txt`
- implemented new error handler for validation errors
- implemented validation mechanism + decorator
- provided three schemas with additional validation and adopted the new decorator on their respective endpoints
- added tests for all new code

## How Has This Been Tested?
- unit tests
- e2e tests
- manually
- postman

## PR Quality Checklist

### DIFF
--- a/api/exception/handlers.py
+++ b/api/exception/handlers.py
@@ -1,6 +1,7 @@
 from boto3.exceptions import Boto3Error
 from botocore.exceptions import ClientError
 from flask import jsonify
+from marshmallow import ValidationError
 from werkzeug.routing import WebsocketMismatch
 
 from api.exception import CSRFError
@@ -35,6 +36,12 @@ def unauthenticated_error_handler(err):
     logger.error(descr, extra=dict(status=code, exception=type(err)))
     return __handler_response(code, descr)
 
+
+def validation_error_handler(err: ValidationError):
+    descr, code = str(err), 400
+    logger.error(descr, extra=dict(status=code, exception=type(err), validation_errors=err.data))
+    return __handler_response(code, descr, validation_errors=err.data)
+
 def global_exception_handler(err):
     try:
         code = err.code
@@ -46,8 +53,8 @@ def global_exception_handler(err):
     return __handler_response(code, 'An error occurred while trying to complete your request. Please try again later. If the problem persists, please contact support for further assistance.')
 
 
-def __handler_response(code, description='Something went wrong'):
-    response = {'code': code, 'message': description}
+def __handler_response(code, description='Something went wrong', **kwargs):
+    response = {'code': code, 'message': description, **kwargs}
     return jsonify(response), code
 
 
@@ -63,6 +70,7 @@ class ExceptionHandler(object):
         app.register_error_handler(ClientError, boto3_exception_handler)
 
         app.register_error_handler(CSRFError, csrf_error_handler)
+        app.register_error_handler(ValidationError, validation_error_handler)
         app.register_error_handler(ValueError, value_error_handler)
         app.register_error_handler(RefreshTokenError, unauthenticated_error_handler)
         app.register_error_handler(Exception, global_exception_handler)
--- a/api/tests/exceptions/test_exception_handlers.py
+++ b/api/tests/exceptions/test_exception_handlers.py
@@ -32,6 +32,16 @@ def test_unauthenticated_error_handler(client, app, monkeypatch):
     assert response.status_code == 401
     assert response.json == {'code': 401, 'message': 'Refresh token error: refresh-token-error'}
 
+def test_validation_error_handler(client, mock_enable_auth, mock_csrf_needed):
+
+    response = client.post('/manager/ec2_action', query_string={'region': 'eu-west-1', 'instance_ids': 'i-1,i-2', 'action': 'invalid-action'})
+
+    assert response.status_code == 400
+    assert response.json == {
+        'code': 400, 'message': 'Input validation failed for /manager/ec2_action',
+        'validation_errors': {'action': ['Must be one of: stop_instances, start_instances.']}
+    }
+
 def test_global_exception_handler_with_app_logic(client, app, monkeypatch):
     def get_app_config_raising_generic_exception():
         raise Exception('generic exception')
--- a/api/tests/exceptions/test_exception_handlers.py
+++ b/api/tests/exceptions/test_exception_handlers.py
@@ -1,3 +1,5 @@
+from marshmallow import ValidationError
+
 from api.exception.exceptions import RefreshTokenError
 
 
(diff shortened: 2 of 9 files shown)


api/validation/schemas.py
```diff
@@ -0,0 +1,26 @@
+from marshmallow import Schema, fields, validate, INCLUDE
+
+from api.validation.validators import comma_splittable, aws_region_validator
+
+
+class EC2ActionSchema(Schema):
+    action = fields.String(required=True, validate=validate.OneOf(['stop_instances', 'start_instances']))
```
mtfranchetto, line 7: question: what's the correct tradeoff between typing a field as string, or enumerate all possible values? In this case the parameter has only two values, so it's fairly easy to do a precise typing, but what about an endpoint which allows, for example, 10 different actions? Do we have to enumerate them all? What's your recommendation?

api/validation/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+    if args_schema:
+        errors.update(args_schema.validate(_request.args))
+
+    if cookies_schema:
+        errors.update(cookies_schema.validate(_request.cookies))
+
+    return errors
+
+
+def valid(*, json: Schema = None, args: Schema = None, cookies: Schema = None):
```
mtfranchetto, line 26: suggestion(non-blocking): to enhance readability, and since this utility is made to validate requests, I would rename `json` to `body` and `args` to `params`, more related to the HTTP jargon

app.py
```diff
@@ -136,12 +139,14 @@ def list_users_():
     @app.route("/manager/create_user", methods=["POST"])
     @authenticated(ADMINS_GROUP)
     @csrf_needed
+    @valid(json=CreateUser)
```
mtfranchetto, line 142: nitpick: I would use `validated` to be consistent with other decorators

api/validation/schemas.py
```diff
@@ -0,0 +1,26 @@
+from marshmallow import Schema, fields, validate, INCLUDE
```
mtfranchetto, line 1: question: since we have to cover all application endpoints with validation, is it safe to have all schemas under a single file?

api/tests/exceptions/test_exception_handlers.py
```diff
@@ -32,6 +32,16 @@ def push_log_raising():
     assert response.status_code == 401
     assert response.json == {'code': 401, 'message': 'Refresh token error: refresh-token-error'}
 
+def test_validation_error_handler(client, mock_enable_auth, mock_csrf_needed):
```
mendaomn, line 35: issue: If possible, avoid relying on the actual implementation of `EC2Action`. This way we can safely change the validation of the `ec2_action` endpoint without breaking this test.

I'd recommend just calling the handler and asserting against outputs, given controlled inputs

api/exception/handlers.py
```diff
@@ -46,8 +53,8 @@ def global_exception_handler(err):
     return __handler_response(code, 'An error occurred while trying to complete your request. Please try again later. If the problem persists, please contact support for further assistance.')
 
 
-def __handler_response(code, description='Something went wrong'):
-    response = {'code': code, 'message': description}
+def __handler_response(code, description='Something went wrong', **kwargs):
```
mendaomn, line 56: question: Are we sure we want to return the input validation errors as they are? What could this information be used for?

api/validation/__init__.py
```diff
@@ -0,0 +1,37 @@
+from functools import wraps
+
+from flask import request, Request
+from marshmallow import Schema, ValidationError
+
+from api.validation.schemas import EC2Action
+
+
+def __valid_request(_request: Request, *, json_schema: Schema = None, args_schema: Schema = None, cookies_schema: Schema = None):
```
mendaomn, line 9: readability: would `validate_request` be a more representative name?

mendaomn:
General questions: how is the text of the error messages generated? It's not a concern for this PR, but I'm wondering:

- can they be centralized, or otherwise searched through in case we ever needed to spot a specific issue starting from the error message? (e.g. we find the error message, and we need to get to the schema definition to see if there is a bug)
- can they be internationalised, or customized at all?

api/validation/__init__.py
BarcoMasile: I like it, but I'd still like to keep it "private" with the double underscore since as of now the function is not intended to be used elsewhere.
Coming up in next revision, thanks!

app.py
BarcoMasile: Nice catch, thanks! Coming up in next revision!

api/validation/__init__.py
BarcoMasile: I was actually indecisive between these and the one you propose, so I'll go with yours (which I actually prefer)
Thanks!

api/validation/schemas.py
BarcoMasile: My personal recommendation is to be as strict as possible, when it makes sense (so without getting crazy).
For example for PC available regions, we know that values will rarely change, so we can handle them all.

I would keep having this approach if values don't change too often and are not too big of a number, and they're static values.

For all other instances we can just type a field as string and let the target service do the rest

api/tests/exceptions/test_exception_handlers.py
BarcoMasile: Right! I'll follow your recommendation sir, thanks!

api/validation/schemas.py
BarcoMasile: I believe it's safe if they're still "readable"/"searchable" easily
But I strongly believe in separating code in packages, so as soon as we find the schemas need better organization/structure we can do the following:
- restructure the schemas in separate files under `api/validation` package
- export them all from the `api/validation/__init__.py` so they're still all importable from `from api.validaton import CertainSchema`

BarcoMasile:
> General questions: how is the text of the error messages generated? It's not a concern for this PR, but I'm wondering:
> 
> * can they be centralized, or otherwise searched through in case we ever needed to spot a specific issue starting from the error message? (e.g. we find the error message, and we need to get to the schema definition to see if there is a bug)
> * can they be internationalised, or customized at all?

The library of choice allows for [customization](https://marshmallow.readthedocs.io/en/stable/custom_fields.html#customizing-error-messages) of error message at the field level so we can (in the future) for example return i18n keys to the frontend and let the frontend translate those for the users
@mendaomn

api/exception/handlers.py
mendaomn: Yeah but validation errors will not be handled by the frontend, will they? Either way, I agree there is not sensitivi info

mendaomn:
> > General questions: how is the text of the error messages generated? It's not a concern for this PR, but I'm wondering:
> > 
> > * can they be centralized, or otherwise searched through in case we ever needed to spot a specific issue starting from the error message? (e.g. we find the error message, and we need to get to the schema definition to see if there is a bug)
> > * can they be internationalised, or customized at all?
> 
> The library of choice allows for [customization](https://marshmallow.readthedocs.io/en/stable/custom_fields.html#customizing-error-messages) of error message at the field level so we can (in the future) for example return i18n keys to the frontend and let the frontend translate those for the users @mendaomn

Good, I noticed you added the `request.path` in the `ValidationError` constructor. Thus, we should be able to easily traceback to the failing schema in case of validation errors. Good job!

PR #477, opened 2023-01-11
Fixed get_identity causing full login processes

BarcoMasile:
## Description
Removed 401 from get_identity + adjusted tests
## How Has This Been Tested?
- manual
- unit tests

## PR Quality Checklist

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -588,9 +588,6 @@ def get_identity():
     try:
         decoded_access = jwt_decode(access_token)
     except jwt.ExpiredSignatureError:
-        if not auth_cookies:
-            return {}, 401
-
         decoded_access = jwt_decode(auth_cookies['accessToken'])
 
     identity = _get_identity_from_token(decoded=decoded_access, claims=claims)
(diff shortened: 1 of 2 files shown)


api/PclusterApiHandler.py
```diff
@@ -582,27 +582,29 @@ def get_identity():
...
 
+    claims = ["email"]
+    try:
         decoded_access = jwt_decode(access_token)
-        identity_from_access_token = _get_identity_from_token(decoded=decoded_access, claims=claims)
+    except jwt.ExpiredSignatureError:
+        if not auth_cookies:
+            return {}, 401
+
+        decoded_access = jwt_decode(auth_cookies['accessToken'])
```
mtfranchetto, line 594: question: why is this fallback mechanism needed?

BarcoMasile: strictly speaking this is not actually needed, since it's taken care of by the `@authenticated` decorator
I put it as a defensive approach and also an explanatory action to the reader
I'm open to remove it !

BarcoMasile: On a second thought, I decided to remove it 
Thanks!

mtfranchetto (dismissed):

mtfranchetto (approved):

PR #479, opened 2023-01-12
Fix "Stop Compute Fleet" button still spinning after the compute fleet is stopped

### LINKED ISSUE #150: Stop Compute Fleet Spinning
Stop Compute Fleet (from the Edit dialogue) never stops spinning even after the cluster stops.

mtfranchetto:
## Description

Introduces an hook to poll the details of a cluster to fix an erroroneus behavior of the "Stop compute fleet" button inside the wizard.
I choose a plain `setInterval` approach instead of going full `react-query` because its refetch API is tricky to use with this specific case.

Fixes #150.

## How Has This Been Tested?

- stopped a cluster and verified that split panel details are updated correctly
- stopped a cluster while editing its configuration and verified the button stops spinning after shutdown

## PR Quality Checklist

### DIFF
--- a/frontend/src/components/__tests__/useClusterPoll.test.tsx
+++ b/frontend/src/components/__tests__/useClusterPoll.test.tsx
@@ -1,0 +1,57 @@
+import {renderHook} from '@testing-library/react-hooks'
+import {useClusterPoll} from '../useClusterPoll'
+
+jest.mock('../../model', () => ({
+  DescribeCluster: jest.fn(),
+}))
+import {DescribeCluster} from '../../model'
+import {act} from 'react-dom/test-utils'
+
+describe('Given a cluster poll', () => {
+  beforeEach(() => jest.useFakeTimers())
+  afterEach(() => {
+    jest.resetAllMocks()
+    jest.useRealTimers()
+  })
+
+  describe('when a cluster name is given', () => {
+    it('should start polling the resource', () => {
+      renderHook(() => useClusterPoll('Test', true))
+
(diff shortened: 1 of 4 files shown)


frontend/src/components/useClusterPoll.tsx
```diff
@@ -0,0 +1,20 @@
+import {useEffect, useState} from 'react'
+import {DescribeCluster} from '../model'
+
+export const useClusterPoll = (clusterName: string, startOnRender: boolean) => {
+  const [start, setStart] = useState(startOnRender)
+
+  useEffect(() => {
+    if (!start) return
```
mendaomn, line 8: issue: should we clear the interval when `start` goes to false after `stop()` has been called?

frontend/src/components/__tests__/useClusterPoll.test.tsx
```diff
@@ -0,0 +1,57 @@
+import {renderHook} from '@testing-library/react-hooks'
```
mendaomn, line 1: praise: great test, it's readable and easy to follow even if it's more complicated than your average test, well done!

frontend/src/components/useClusterPoll.tsx
BarcoMasile: isn't this already taken care of of the code at line 13 ?

mtfranchetto: Yes, it should be handled already by L13, otherwise I haven't understood your point

BarcoMasile (approved):

PR #482, opened 2023-01-12
Removed print statements from the codebase

BarcoMasile:
Removed print statements from the codebase.
Left them in the infrastructure file since they could be useful there.

### DIFF
--- a/api/PclusterApiHandler.py
+++ b/api/PclusterApiHandler.py
@@ -346,22 +346,17 @@ def sacct():
         if isinstance(accounting, tuple):
             return accounting
         # Try to retrieve relevant cost information
-        try:
-            queue_name = json.loads(accounting)[0]["partition"]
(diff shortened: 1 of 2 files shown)


api/PclusterApiHandler.py
```diff
@@ -360,7 +357,7 @@ def sacct():
             if not isinstance(_price_guess, tuple):
                 price_guess = _price_guess
         except Exception as e:
-            print(e)
+            pass
```
tmscarla, line 360: thought: shall we log the exception in this case?

tmscarla (approved):

api/PclusterApiHandler.py
BarcoMasile: We could probably raise the exception and let our handlers deal with that (log and return the error)
Yes I'll do that

tmscarla (approved):

PR #484, opened 2023-01-13
Fix logger pushing non-string payloads to remote logging system

mendaomn:
## Description

This PR fixes an issue allowing any payload to be sent as message to the remote logging system. 

This is mostly related to the fact that rejected promises can contain any value, thus by blindly logging the `event.reason` of the `PromiseRejectionEvent` we risk to push non-string values to the remote logging system

This could also occur in the future, by logging values that have been typed as `any` (which would defeat the type checking), therefore the guard is place inside the `RemoteLogger`

## Changes

- log stacktraces when Errors are logged
- stringify non-string messages before pushing them
- handle undefined messages

## How Has This Been Tested?

- manually
- unit tests

Log entry to be sent to the remote logging system, in case of an uncaught axios error:
image

## PR Quality Checklist

### DIFF
--- a/frontend/src/logger/ILogger.ts
+++ b/frontend/src/logger/ILogger.ts
@@ -9,7 +9,11 @@ export interface ILogger {
 
   debug(message: string, extra?: Record<string, unknown>, source?: string): void
 
-  error(message: string, extra?: Record<string, unknown>, source?: string): void
+  error(
+    message: Error | string,
+    extra?: Record<string, unknown>,
+    source?: string,
+  ): void
 
   critical(
     message: string,
--- a/frontend/src/logger/ILogger.ts
+++ b/frontend/src/logger/ILogger.ts
@@ -16,7 +16,7 @@ export interface ILogger {
   ): void
 
   critical(
-    message: string,
+    message: Error | string,
     extra?: Record<string, unknown>,
     source?: string,
   ): void
--- a/frontend/src/logger/RemoteLogger.ts
+++ b/frontend/src/logger/RemoteLogger.ts
@@ -116,7 +116,7 @@ export class Logger implements ILogger {
     if (payload instanceof Error) {
       extra = {
         ...extra,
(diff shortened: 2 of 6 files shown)


frontend/src/logger/RemoteLogger.ts
```diff
@@ -20,6 +20,8 @@ type BufferConfig = {
   window: number
 }
 
+const DEAFULT_MESSAGE = 'This log entry has no message.'
```
mtfranchetto, line 23: todo: small typo!

frontend/src/logger/ILogger.ts
```diff
@@ -9,7 +9,11 @@ export interface ILogger {
...
   debug(message: string, extra?: Record<string, unknown>, source?: string): void
 
-  error(message: string, extra?: Record<string, unknown>, source?: string): void
+  error(
+    message: Error | string,
+    extra?: Record<string, unknown>,
+    source?: string,
+  ): void
 
   critical(
```
mtfranchetto, line 18: question: shall we pass errors to `critical` as well?

frontend/src/logger/RemoteLogger.ts
```diff
@@ -96,9 +102,27 @@ export class Logger implements ILogger {
...
   ): LogEntry {
-    return {message, level, extra}
+    if (!payload) {
+      return {message: DEAFULT_MESSAGE, level, extra}
+    }
+
+    if (payload instanceof Error) {
+      extra = {
+        ...extra,
+        trace: payload.stack,
```
mtfranchetto, line 115: suggestion(non-blocking): I'd call this field `stacktrace` since trace is vague and makes me think of tracing

frontend/src/logger/ILogger.ts
mendaomn: makes sense to me

mtfranchetto (approved):

PR #485, opened 2023-01-13
Improved validation constraints for API signatures

BarcoMasile:
## Description
Added size constraints to API signatures validation

## How Has This Been Tested?
- manually
- all the test suits

## PR Quality Checklist

### DIFF
--- a/api/validation/schemas.py
+++ b/api/validation/schemas.py
@@ -1,12 +1,10 @@
 from marshmallow import Schema, fields, validate, INCLUDE
 
 from api.validation.validators import comma_splittable, aws_region_validator, is_alphanumeric_with_hyphen, valid_api_log_levels_predicate
-from api.logging import VALID_LOG_LEVELS
-
 
 class EC2ActionSchema(Schema):
     action = fields.String(required=True, validate=validate.OneOf(['stop_instances', 'start_instances']))
-    instance_ids = fields.String(required=True, validate=comma_splittable)
+    instance_ids = fields.String(required=True, validate=validate.And(comma_splittable, validate.Length(max=2048)))
     region = fields.String(validate=aws_region_validator)
 
 
@@ -14,8 +12,8 @@ EC2Action = EC2ActionSchema(unknown=INCLUDE)
 
 
 class CreateUserSchema(Schema):
-    Username = fields.Email(required=True)
-    Phonenumber = fields.String()
+    Username = fields.Email(required=True, validate=validate.Length(max=320)) # Email RFC allows max 320 chars
+    Phonenumber = fields.String(validate=validate.Length(max=15)) # ITU-T E.164 allows phone numbers no more than 15 digits
 
 
 CreateUser = CreateUserSchema(unknown=INCLUDE)
@@ -29,13 +27,13 @@ DeleteUser = DeleteUserSchema(unknown=INCLUDE)
 
 class GetClusterConfigSchema(Schema):
     region = fields.String(validate=aws_region_validator)
-    cluster_name = fields.String(required=True, validate=is_alphanumeric_with_hyphen)
+    cluster_name = fields.String(required=True, validate=validate.And(is_alphanumeric_with_hyphen, validate.Length(max=60))) # PC allow cluster name of max 60 chars
 
 GetClusterConfig = GetClusterConfigSchema(unknown=INCLUDE)
 
 
 class GetCustomImageConfigSchema(Schema):
-    image_id = fields.String(required=True, validate=is_alphanumeric_with_hyphen)
+    image_id = fields.String(required=True, validate=validate.And(is_alphanumeric_with_hyphen, validate.Length(min=1, max=1024))) # AMI id min 1, max 1024 chars
 
 GetCustomImageConfig = GetCustomImageConfigSchema(unknown=INCLUDE)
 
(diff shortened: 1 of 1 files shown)


lazyoft (commented):
Shouldn't the test reflect the newly introduced constraints somehow?

api/validation/schemas.py
```diff
@@ -1,21 +1,19 @@
...
     region = fields.String(validate=aws_region_validator)
 
 
 EC2Action = EC2ActionSchema(unknown=INCLUDE)
 
 
 class CreateUserSchema(Schema):
-    Username = fields.Email(required=True)
-    Phonenumber = fields.String()
+    Username = fields.Email(required=True, validate=validate.Length(max=320)) # Email RFC allows max 320 chars
```
lazyoft, line 15: praise: I love the references here

api/validation/schemas.py
```diff
@@ -53,72 +51,72 @@ class GetInstanceTypesSchema(Schema):
...
 
 CancelJob = CancelJobSchema(unknown=INCLUDE)
 
 
 class SacctSchema(Schema):
-    user = fields.String()
-    instance_id = fields.String(required=True)
-    cluster_name = fields.String(required=True, validate=is_alphanumeric_with_hyphen)
+    user = fields.String(validate=validate.Length(max=64))
+    instance_id = fields.String(required=True, validate=validate.Length(max=32))
```
lazyoft, line 89: question: is the instance_id for DCV referring to a different object? If not, why in one case I have 60 and in this case 32?

api/validation/schemas.py
```diff
@@ -53,72 +51,72 @@ class GetInstanceTypesSchema(Schema):
...
+    code = fields.String(required=True, validate=validate.Length(max=128))
 
 Login = LoginSchema(unknown=INCLUDE)
 
 
 class PushLogSchema(Schema):
     class PushLogEntrySchema(Schema):
         level = fields.String(required=True, validate=valid_api_log_levels_predicate)
-        message = fields.String(required=True)
+        message = fields.String(required=True, validate=validate.Length(max=246000)) # CW limit is 256k, leaving 1k to extra and level
```
lazyoft, line 105: praise: nice explanation in the comment

api/validation/schemas.py
```diff
@@ -53,72 +51,72 @@ class GetInstanceTypesSchema(Schema):
...
-    queue_name = fields.String(required=True)
+    cluster_name = fields.String(required=True, validate=validate.Length(max=60))
+    queue_name = fields.String(required=True, validate=validate.Length(max=60))
     region = fields.String(validate=aws_region_validator, required=True)
 
 PriceEstimate = PriceEstimateSchema(unknown=INCLUDE)
 
 class PCProxySchema(Schema):
-    path = fields.String(required=True)
+    path = fields.String(required=True, validate=validate.Length(max=512))
```
lazyoft, line 120: question: is this for the YAML configuration? If so I think 512 is too low

BarcoMasile: This is for the path to use to invoke the PC API
512 is large in this case but I put 512 since I have no way of anticipating how long an URL for PC APIs will be
I believe it could be 256 and we'd still be safe

api/validation/schemas.py
BarcoMasile: Thank you!

api/validation/schemas.py
BarcoMasile: My bad, I'll correct it and re-push! Thanks

api/validation/schemas.py
BarcoMasile: Thanks!

BarcoMasile:
> question: shouldn't the test reflect the newly introduced constraints somehow?

I'd say no since validators are part of the third library so since we have tests proving our mechanism (that relies on Marshmallow) works, then it's ok to assume they did their job also

lazyoft:
> > question: shouldn't the test reflect the newly introduced constraints somehow?
> 
> I'd say no since validators are part of the third library so since we have tests proving our mechanism (that relies on Marshmallow) works, then it's ok to assume they did their job also

You are right. I'm ok with these changes.

lazyoft (approved):

api/validation/schemas.py
```diff
@@ -1,21 +1,19 @@
 from marshmallow import Schema, fields, validate, INCLUDE
 
 from api.validation.validators import comma_splittable, aws_region_validator, is_alphanumeric_with_hyphen, valid_api_log_levels_predicate
-from api.logging import VALID_LOG_LEVELS
-
 
 class EC2ActionSchema(Schema):
     action = fields.String(required=True, validate=validate.OneOf(['stop_instances', 'start_instances']))
-    instance_ids = fields.String(required=True, validate=comma_splittable)
+    instance_ids = fields.String(required=True, validate=validate.And(comma_splittable, validate.Length(max=2048)))
```
mtfranchetto, line 7: suggestion: even though the endpoint can receive a list of ids, the UI is configured to operate on a single instance a t a time. I suggest to limit this field to the length of an instance id, and, if you want to do the extra mile, make this endpoint accept a single instance instead of an array

api/validation/schemas.py
```diff
@@ -53,72 +51,72 @@ class GetInstanceTypesSchema(Schema):
...
 class PriceEstimateSchema(Schema):
-    cluster_name = fields.String(required=True)
-    queue_name = fields.String(required=True)
+    cluster_name = fields.String(required=True, validate=validate.Length(max=60))
+    queue_name = fields.String(required=True, validate=validate.Length(max=60))
     region = fields.String(validate=aws_region_validator, required=True)
 
 PriceEstimate = PriceEstimateSchema(unknown=INCLUDE)
 
 class PCProxySchema(Schema):
```
mtfranchetto, line 119: issue(possibly-blocking): with a schema like that a user is still allowed to use a huge payload proxied to the ParallelCluster API. Since we don't know beforehand the body content of mutations routed to PC, can we at least put a hard limit on the bytes of the body?

api/validation/schemas.py
BarcoMasile: Nice one! I'll do it gladly, but I would do this in a new PR so the endpoint (and its validation schema) get changed in the same PR context

api/validation/schemas.py
BarcoMasile: I will investigate this, I think it's an important issue. Thanks!

api/validation/validators.py
```diff
@@ -33,3 +35,9 @@ def is_alphanumeric_with_hyphen(arg: str):
 
 def valid_api_log_levels_predicate(loglevel):
     return loglevel.lower() in VALID_LOG_LEVELS
+
+def size_not_exceeding(data, size):
```
mtfranchetto, line 39: todo: don't forget to add some tests on the new validator and on the `raise_on_missing_body` flag as well

mtfranchetto (approved):

PR #488, opened 2023-01-17
Adopted sha256 for CSRF token generation and improved CSRFError handler + tests

BarcoMasile:
## Description
This PR takes care of three things:
- switches to sha256 for the strategy used when creating a CSRF token
- in case of a CSRF token error the cookie gets deleted
- tests are adjusted and added for csrf exception handler

## How Has This Been Tested?
- unit tests
- e2e
- manual tests running "state changing" requests that trigger the csrf token check

## PR Quality Checklist

### DIFF
--- a/api/security/csrf/csrf.py
+++ b/api/security/csrf/csrf.py
@@ -9,16 +9,18 @@ from api.exception import CSRFError
 from api.security.csrf import CSRF_SECRET_KEY
 from api.security.csrf.constants import CSRF_COOKIE_NAME, SALT, CSRF_TOKEN_HEADER
 
+digest_method = staticmethod(hashlib.sha256)
+signer_kwargs = {'signer_kwargs': {'digest_method': digest_method}}
 
 def generate_csrf_token(secret_key, salt):
-    serializer = URLSafeSerializer(secret_key, salt)
+    serializer = URLSafeSerializer(secret_key, salt, **signer_kwargs)
     csrf_token_value = hashlib.sha256(os.urandom(64)).hexdigest()
     csrf_token = serializer.dumps(csrf_token_value)
     return csrf_token
 
 
 def parse_csrf_token(secret_key, salt, token):
-    serializer = URLSafeSerializer(secret_key, salt)
+    serializer = URLSafeSerializer(secret_key, salt, **signer_kwargs)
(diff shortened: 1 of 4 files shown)


api/security/csrf/csrf.py
```diff
@@ -39,6 +41,8 @@ def is_csrf_enabled():
 def set_csrf_cookie(resp, csrf_token):
     resp.set_cookie(CSRF_COOKIE_NAME, value=csrf_token, httponly=True, secure=True, samesite="Lax")
 
+def remove_csrf_cookie(resp):
+    resp.set_cookie(CSRF_COOKIE_NAME, value='', expires=0)
```
mendaomn, line 45: question: is this better than [delete_cookie](https://flask.palletsprojects.com/en/2.2.x/api/#flask.Response.delete_cookie)?

api/tests/exceptions/test_exception_handlers.py
```diff
@@ -23,6 +24,23 @@ def push_log_raising():
     assert response.status_code == 400
     assert response.json == {'code': 400, 'message': 'Validation error'}
 
+def test_csrf_error_exception_handler(client, app, monkeypatch):
+
+    def sacct_raising():
+        raise CSRFError('CSRF Error')
+
+    monkeypatch.setitem(app.view_functions, 'sacct_', sacct_raising)
+    response = client.post('/manager/sacct')
```
mendaomn, line 33: issue: let's try to avoid binding tests to specific endpoints, when reasonable. This way we don't have to update this test in case that endpoint changes or is removed

In this case you are testing the handler, I would call the handler directly with a CSRFError and assert on the returned response

api/security/csrf/csrf.py
BarcoMasile: It's not better, I didn't know the `delete_cookie` function. I'll switch to that, nice!

api/tests/exceptions/test_exception_handlers.py
BarcoMasile: I like the proposal, although I would have to refactor all the endpoints (and I wouldn't do it in this PR)
I'll change the current one and take notes!

mendaomn (approved):

PR #495, opened 2023-01-20
Improved CSRF token handling with TimedSignature

BarcoMasile:
## Description
This PR changes the way CSRF token (and cookie) are generated, moving from a `URLSafeSerializer` to a `URLSafeTimedSerializer`.
Now CSRF tokens are valid for 30 seconds, after that they would throw bad signature exception.
Now invocation of the `/csrf` token always returns (and sets as an expiring cookie) a new token.

## How Has This Been Tested?
- unit tests
- e2e
- manual tests

## PR Quality Checklist

### DIFF
--- a/api/tests/security/csrf/test_csrf.py
+++ b/api/tests/security/csrf/test_csrf.py
@@ -1,3 +1,4 @@
+import datetime
 import time
 from unittest.mock import ANY, call
 
@@ -29,6 +30,8 @@ def test_crsf_extension_get_new_csrf(app):
               it should return the csrf_token in a json
     """
     CSRF(app)
+    now = datetime.datetime.utcnow()
+    expected_expiration = (now + datetime.timedelta(seconds=30)).strftime('%a, %d %b %Y %H:%M:%S')
     resp = app.test_client().get('/csrf')
     csrf_cookies = list(cookie_value for cookie_header, cookie_value in resp.headers if
                         'Set-Cookie' in cookie_header and CSRF_COOKIE_NAME in cookie_value)
@@ -37,7 +40,7 @@ def test_crsf_extension_get_new_csrf(app):
     assert 'csrf_token' in resp.json
     assert len(csrf_cookies) > 0
     assert 'Secure; HttpOnly; Path=/; SameSite=Lax' in csrf_cookies[0]
-    assert 'Expires=' in csrf_cookies[0]
+    assert f'Expires={expected_expiration}' in csrf_cookies[0]
 
 
 def test_csrf_needed_decorator(mocker, capsys, mock_csrf_token_string, app, mock_parse_csrf):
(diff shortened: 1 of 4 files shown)


mtfranchetto (dismissed):

api/tests/security/csrf/test_csrf.py
```diff
@@ -36,27 +37,7 @@ def test_crsf_extension_get_new_csrf(app):
...
-    resp = test_client.get('/csrf')
-    csrf_cookies = list(cookie_value for cookie_header, cookie_value in resp.headers if
-                        'Set-Cookie' in cookie_header and CSRF_COOKIE_NAME in cookie_value)
-
-    assert resp.status_code != 405
-    assert 'csrf_token' in resp.json
-    assert resp.json.get('csrf_token') == 'csrf-value'
-    assert len(csrf_cookies) > 0
-    assert 'Secure; HttpOnly; Path=/; SameSite=Lax' in csrf_cookies[0]
+    assert 'Expires=' in csrf_cookies[0]
```
mendaomn, line 40: issue: let's assert the cookie actually expires in 30 seconds (by the way, I believe you could set `max_age` instead of `expires` if that helps)

api/tests/security/csrf/test_csrf.py
```diff
@@ -117,6 +98,12 @@ def test_parse_csrf_token(mock_urandom, mock_csrf_token_value):
 
     assert actual_token_value == mock_csrf_token_value
 
+def test_parse_csrf_token_expired(mock_urandom):
+    csrf_token = generate_csrf_token(SECRET_KEY,SALT)
+    time.sleep(1)
+
+    with pytest.raises(BadSignature):
+        parse_csrf_token(SECRET_KEY, SALT, csrf_token, max_age=0)
```
mendaomn, line 106: question: how does this work? validation fails if the signature is older than max_age=0? can we use a mechanism other than using an actual sleep? (like fake timers)

BarcoMasile: I already looked for fake timers but thery're not included in the pytest suit, so we would need another dependency and I thought that for this single instance it was not worth it

lazyoft: I would inject a timeprovider or something similar that can be mocked, and use that to request the time on the parse function.

BarcoMasile: Unfortunately there's not one out of the box. We would need another library like [freezegun](https://github.com/spulec/freezegun)

api/tests/security/csrf/test_csrf.py
BarcoMasile: coming up in next rev

mendaomn (approved):

PR #496, opened 2023-01-20
Retrieve CSRF token before each mutation

mtfranchetto:
## Description

Align the frontend after the backend changes for the new CSRF token generation mechanism #495 

## Changes
- add the ability to pass headers to the `executeRequest` function
- moved the CSRF token retrieval from the initial state to the `executeRequest` function
- stub `executeRequest` instead of `axios` in `CreateCluster` tests, otherwise the CSRF check complains

## How Has This Been Tested?

Run the portal locally and verified:
- no CSRF call is made at the startup
- a CSRF token is retrieved before each mutation

## PR Quality Checklist


frontend/src/model.tsx
```diff
@@ -899,24 +895,10 @@ async function GetAppConfig() {
...
 
-async function LoadCsrfToken(logger: ILogger): Promise<string | null> {
-  try {
-    const token = await getCsrfToken(axiosInstance)
-    setCsrfTokenHeader(axiosInstance, token)
-    logger.info(`Set X-CSRF-Token header to ${token}`)
-    return token
-  } catch (error) {
-    logger.error('Could not fetch CSRF token')
-    notify(i18n.t('global.errors.csrfError'), 'error')
```
mendaomn, line 910: nit-pick: we can get rid of `global.errors.csrfError`

BarcoMasile (dismissed):

frontend/src/http/__tests__/csrf.test.ts
```diff
@@ -0,0 +1,72 @@
...
+          body: 'test',
+        })
+
+        expect(data).toMatchObject({
+          success: true,
+        })
+      })
+    })
+
+    describe('when a CSRF token request fails', () => {
```
mendaomn, line 59: question: double checking I got this right:
- the caller will handle the error
- logging is delegated to the axios interceptors

is that correct?

mtfranchetto: Correct!

frontend/src/model.tsx
mtfranchetto: Right!

mtfranchetto: Removed

mendaomn (approved):

PR #511, opened 2023-01-25
Update copy of Users page

mendaomn:
## Description

This PR updates copy of Users page, and adds hardcoded strings to our localization file

## How Has This Been Tested?

- manually

## PR Quality Checklist

### DIFF
--- a/frontend/locales/en/strings.json
+++ b/frontend/locales/en/strings.json
@@ -139,7 +139,10 @@
         "description": "The list of your clusters."
       },
       "dialogs": {
-        "delete": "Are you sure you want to delete cluster {{clusterName}}?"
+        "delete": {
+          "title": "Delete cluster",
+          "body": "Are you sure you want to delete cluster {{clusterName}}?"
+        }
       },
       "actions": {
         "create": "Create",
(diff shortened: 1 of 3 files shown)


frontend/locales/en/strings.json
```diff
@@ -583,6 +586,50 @@
       }
     }
   },
+  "users": {
+    "header": {
+      "title": "Users",
+      "description": "The list of users that have permission to access the AWS ParallelCluster User Interface."
+    },
+    "actions": {
+      "create": "Create user",
```
tmscarla, line 595: thought: we should rely on global actions where possible. If I am not mistaken `refresh` and `delete` should be already defined.

frontend/locales/en/strings.json
```diff
@@ -583,6 +586,50 @@
...
+        }
+      },
+      "createForm": {
+        "phoneNumberPlaceholder": "+11234567890",
+        "emailPlaceholder": "email@domain.com"
+      },
+      "filtering": {
+        "loadingText": "Loading users...",
+        "countText": "Results: {{filteredItemsCount}}",
+        "filteringAriaLabel": "Enter search text",
```
tmscarla, line 619: praise: nice catch!

frontend/locales/en/strings.json
```diff
@@ -583,6 +586,50 @@
...
+        "filteringPlaceholder": "Enter search text",
+        "empty": {
+          "title": "No users",
+          "subtitle": "No users to display."
+        },
+        "noMatch": {
+          "title": "No matches",
+          "subtitle": "No users match the filters.",
+          "action": "Clear filter"
```
tmscarla, line 628: thought: same as above. This action should be the same for all tables in which we use filtering.

tmscarla (approved):

frontend/locales/en/strings.json
mendaomn: It's a good point. I thought about it and my conclusion was that we should be sure the two actions will always change copy "together". In general, re-using keys is a bit troublesome. We could maybe create a Component to re-use everywhere, thus we'll have only one key.

I propose we do it in future PRs

PR #541, opened 2023-02-01
Add AbsoluteTimestamp component to display dates following Cloudscape guidelines

tmscarla:
## Description
* Add `AbsoluteTimestamp` component to display dates following Cloudscape guidelines

## How Has This Been Tested?
* Manually on local environment

## References

* https://cloudscape.design/patterns/general/timestamps/

## PR Quality Checklist

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -61,6 +61,7 @@
     "jest-mock-extended": "^2.0.6",
     "lint-staged": "^13.0.3",
     "prettier": "^2.7.1",
+    "timezone-mock": "^1.3.6",
     "typescript": "^4.7.4"
   },
   "lint-staged": {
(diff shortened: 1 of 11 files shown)


frontend/package.json
```diff
@@ -24,7 +24,7 @@
   "scripts": {
     "start": "next start",
     "build": "next build",
-    "test": "jest",
+    "test": "TZ=utc jest",
```
mtfranchetto, line 27: question: why are we forcing this timezone for all tests? Isn't it better to use something like `timezone-mock`?

tmscarla: The issue is that being `Timezone.Local` the default option if not timezone is provided, tests were passing on local environment and failing on the CI (due to different locales).

I am open to suggestions. Going to check `timezone-mock`.

mtfranchetto: Yes, I know the issue and why you added `TZ=utc`, I just think it's better to mock it case by case than globally

mtfranchetto (approved):

PR #548, opened 2023-02-03
Fix cluster instances tab copy and action buttons

tmscarla:
## Description

* Fix copywriting for instance stop action button
* Make the buttons always visible but disabled if no instances is selected following Cloudscape guidelines

## How Has This Been Tested?

* Manually on local environment

## References

* https://cloudscape.design/components/button/

## Screenshots

Screenshot 2023-02-03 at 12 30 21

## PR Quality Checklist

### DIFF
--- a/frontend/src/old-pages/Clusters/Instances.tsx
+++ b/frontend/src/old-pages/Clusters/Instances.tsx
@@ -16,15 +16,8 @@ import {useNavigate} from 'react-router-dom'
 import {useTranslation} from 'react-i18next'
 
 import {GetClusterInstances, Ec2Action} from '../../model'
-import {
-  setState,
-  clearState,
-  useState,
-  getState,
-  consoleDomain,
-} from '../../store'
+import {useState, getState, consoleDomain} from '../../store'
 
-// UI Elements
 import {
   Button,
   Header,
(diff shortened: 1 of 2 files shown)


frontend/src/old-pages/Clusters/Instances.tsx
```diff
@@ -42,77 +42,94 @@ import {InstanceStatusIndicator} from '../../components/Status'
...
+      instance.state === InstanceState.Running,
+    [instance],
+  )
+  const isHeadNodeStopped = React.useMemo(
+    () =>
+      instance &&
+      instance.nodeType === NodeType.HeadNode &&
+      instance.state === InstanceState.Stopped,
+    [instance],
+  )
```
mendaomn, line 108: suggestion: useMemo is not needed here for two reasons:

- the value is not passed as props to a Component
- it is a boolean value, thus its reference does not change across re-renders

frontend/src/old-pages/Clusters/Instances.tsx
```diff
@@ -42,77 +42,94 @@ import {InstanceStatusIndicator} from '../../components/Status'
...
-              <Button disabled={true}>
-                {t('cluster.instances.actions.stop')}
-              </Button>
-            )}
-        </div>
-      )}
+      <Button
+        disabled={!(instance && isComputeFleetStopped && isHeadNodeRunning)}
+        loading={stopPending}
+        onClick={() => {
```
mendaomn, line 115: suggestion: if you want to make use of your `useCallback`s, you should either memo-ize this function as well, or pass the `stopInstance` directly (same goes for `startInstance`)

frontend/src/old-pages/Clusters/Instances.tsx
```diff
@@ -42,77 +42,94 @@ import {InstanceStatusIndicator} from '../../components/Status'
...
 
   const refresh = () => {
     const clusterName = getState(['app', 'clusters', 'selected'])
     clusterName &&
-      GetClusterInstances(clusterName, () =>
-        clearState(['app', 'clusters', 'action', 'pending']),
-      )
+      GetClusterInstances(clusterName, () => {
+        resetPending
+      })
```
mendaomn, line 72: issue: since `resetPending` is never called, the button will always be stuck in a loading state

frontend/src/old-pages/Clusters/Instances.tsx
```diff
@@ -140,12 +157,6 @@ export default function ClusterInstances() {
     [],
   )
 
-  const clusterPath = ['clusters', 'index', clusterName]
-  const fleetStatus: ComputeFleetStatus = useState([
-    ...clusterPath,
-    'computeFleetStatus',
-  ])
-
   const onSelectionChangeCallback = React.useCallback(({detail}) => {
```
mendaomn, line 160: issue: not related to this PR, but we noticed that since `instances` is continuously fetched from the backend, it's content changes, but the `selectedInstances` keeps a reference to the old/stale value

for instance, if the istance in the `instances` array goes from status=Running to status=Stopped, this change is never reflected in the `selectedInstance`

frontend/src/old-pages/Clusters/Instances.tsx
```diff
@@ -42,77 +42,94 @@ import {InstanceStatusIndicator} from '../../components/Status'
...
+
+  const clusterName = useState(['app', 'clusters', 'selected'])
+  const fleetStatus: ComputeFleetStatus = useState([
+    'clusters',
+    'index',
+    clusterName,
+    'computeFleetStatus',
+  ])
+  const logHref = `/clusters/${clusterName}/logs?instance=${
+    instance && instance.instanceId
```
mendaomn, line 57: suggestion: even though it is not a problem in this specific case, I'd avoid building links with `null` inside. This evaluates to the string `"null"` when `instance` is null.

A bit safer could be 

```
instance ? `/cluster/...${instance.instanceId}` : ''`
```

frontend/src/old-pages/Clusters/Instances.tsx
tmscarla: Thanks, will track this issue in a separate task.

frontend/src/old-pages/Clusters/Instances.tsx
tmscarla: Thanks, going to remove

